PR #2, opened 2020-04-27
"Extreme" zoom cull data fix

ozlb:
if cull enabled, BB_Grid does not contains points, check if line segment (between points) intersects rectangle.

In the "markup" screenshot, line between 2 red points was not visible before, since points are not contained by gp.BB_Grid.

cull_fix

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -183,6 +183,7 @@ struct ImPlotItem {
     ImPlotItem();
     ~ImPlotItem() { ID = 0; }
     bool Show;
+	bool Highlight;
     ImVec4 Color;
     int NameOffset;
     bool Active;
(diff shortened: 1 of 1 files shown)


epezent:
Awesome. I considered this approach too, but was worried about the performance impact it would have. Before merging, I need to tidy up and optimize the code a bit, then create a add a bench marking test so that we can have better insight into the performance trade offs.

ozlb:
ImRect ClipWith() and Overlaps() they look to me very optimized, so i don't expect it could have impact, but it's a must to have a bench marking test.

ozlb:
Future remark: if BB_Grid does not contains points, check if line segment (between points) intersects rectangle.

PR #3, opened 2020-04-27
Highlight legend overed item

ozlb:
Item legend overed by mouse will highlight the line.
In the example in screenshot, mouse over data2 (lime) highlighted doubling the line weight and giving a color to the label.
highlight

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -183,6 +183,7 @@ struct ImPlotItem {
     ImPlotItem();
     ~ImPlotItem() { ID = 0; }
     bool Show;
+	bool Highlight;
     ImVec4 Color;
     int NameOffset;
     bool Active;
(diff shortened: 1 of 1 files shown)


epezent:
Cool feature! I had to add this manually, so to avoid the changes related to culling (wasn't sure how to selectively merge only the relevant lines). I also made it an optional feature (enabled by default) which can be toggled with ImPlotFlags_Highlight.

epezent:
One thing to note: this doesn't currently affect items that are not lines. We might need to consider how markers/bars should be highlighted (e.g. double their outline weight, or show the outline if it is not visible)

ozlb:
I was thinking about filling with colAlpha but not the outline (if possible) and yes, if outline is not visible showing it)

PR #6, opened 2020-04-29
PlotDigital

ozlb:
PlotDigital


epezent:
This is looking really good. I should have some time today to review.

ozlb:
why new version this is not working anymore...

ozlb:
I update PlotDigital according latest updates and added beta Cursors handling. I would like to improve in order that with x axis zoom change the cursor will keep same delta time.
Peek 2020-05-04 11-23

ozlb:
I don't know if you are already aware of the following "cosmetics":
```
src/implot.cpp: In function ‘void ImGui::EndPlot()’:
src/implot.cpp:1240:16: warning: unused variable ‘flip_x’ [-Wunused-variable]
     const bool flip_x     = HasFlag(plot.XAxis.Flags, ImAxisFlags_Invert);
                ^~~~~~
src/implot.cpp:1245:16: warning: unused variable ‘flip_y’ [-Wunused-variable]
     const bool flip_y     = HasFlag(plot.YAxis.Flags, ImAxisFlags_Invert);
                ^~~~~~

src/implot_demo.cpp: In function ‘void ImGui::ShowImPlotDemoWindow(bool*)’:
src/implot_demo.cpp:221:65: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
         static char* labels1[]   = {"Frogs","Hogs","Dogs","Logs"};
                                                                 ^
src/implot_demo.cpp:221:65: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
src/implot_demo.cpp:221:65: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
src/implot_demo.cpp:221:65: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
src/implot_demo.cpp:242:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
         static char* labels2[]   = {"One","Two","Three","Four","Five"};
                                                                      ^
src/implot_demo.cpp:242:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
src/implot_demo.cpp:242:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
src/implot_demo.cpp:242:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
src/implot_demo.cpp:242:70: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]

```

epezent:
Thanks for bringing this up to speed with the newest optimizations. Are you ready for me to merge this, or is it still WIP?

epezent:
What is "Cursors"?

epezent:
Cursors looks interesting! I would suggest you call it something different, to avoid confusion with mouse cursors. Perhaps "Guide". How are they created? Should there be horizontal guides?

epezent:
Also, do they do anything that the Query cannot already?

epezent:
I opened the discussion; you should be able to comment here now.

ozlb:
> What if instead we made the edges and corners of Query dragable/resizable after it is created? If the query was initially created using the Shift key (to expand vertical), then it would be the exact same behavior as far as I can tell. It also has the added benefit of working in the Y direction when needed. I'd just rather not have two similar tools to manage and upkeep.

Yes, I understand your point. If query expansion (with Shift) will extend Y to +/- INF and have Query resizable (borders not corners, in this way also expanded will work properly), it will for sure be ok. You know, both selections and query selection, they will be fantastic if during selection to see projections on axis (like screenshot markup) and eventually why not also in corner also the size.
Untitled

I will remove XQuery so you can merge PlotDigital

epezent:
I merged this and played around with it for a bit. I like it! A couple suggestions (submit in another PR when you have time).

1. Either use `bool*` for `ys`, OR continue to use `float*` but make the value scale the bit height (e.g., if I pass 2.0f instead of expected 1.0f, I get a double bit height). I think the latter could have some interesting uses. My preference would be to actually implement both a `bool*` and `float*` interface, having the `bool*` version just pass through to the `float`* version with 0.0f and 1.0f y values under the hood.

2. I'd like to think about how we can generalize the concept of freezing data from scroll/zoom, as you have done here for the y-axis. As discussed, I see why this is necessary in this particular application, but I still find it strange that `PlotDigital` is the only API function that does it. I was wondering if we could make it behave as the other functions do (i.e. respond to y drag and zoom by default), but provide an optional function to freeze data along an axis. For example, something like this:

```cpp
ImGui::FreezeYAxis();
ImGui::PlotDigital("Foo",xs,ys,100);
ImGui::UnfreezeYAxis();
```

It could also be a flag setting that can be pushed/popped. But essentially the idea is that in a frozen state, data is interpreted in pixel units, not plot units. This would allow you to do exactly what you want with PlotDigital, but make the API more consistent and also flexible since it could be used on other plot types. Thoughts?

ozlb:
Ok, i will try to follow your indications as you have a better overall vision.

epezent:
If you want to address the first item, I can start to think about the second. I have a general idea of how it should be done.

ozlb:
>     1. Either use `bool*` for `ys`, OR continue to use `float*` but make the value scale the bit height (e.g., if I pass 2.0f instead of expected 1.0f, I get a double bit height). I think the latter could have some interesting uses. My preference would be to actually implement both a `bool*` and `float*` interface, having the `bool*` version just pass through to the `float`* version with 0.0f and 1.0f y values under the hood.
> 
I did some trials, but basically the weak point is that variable bit height will not guarantee to not overlap signals that is basic for this kind of visualization. This is the reason why I add a "common" ImPlotStyleVar_DigitalBitHeight in order to have predefined plot range for each channel.
With variable height, or plots start to jump up and down in plot based on previous or they will overlap.
To be honest probably i didn't catch what you really mean. Probably the answer is to "publish" also the plot offset, so you can do multiple PlotDigital with different ImPlotStyleVar_DigitalBitHeight !?

[PlotDigitalVariableHeight.mov.zip](https://github.com/epezent/implot/files/4597017/PlotDigitalVariableHeight.mov.zip)

ozlb:
What do you think? Check how space reserved to avoid overlap of signals is assigned "dynamically" (i.e. on data2 and data4)
Untitled

epezent:
Looks good! I agree that there should be style variables for both the default size (at y=1) and the padding gap.

PR #14, opened 2020-05-08
Fix MSVC warnings, change getter signature

sayan1an:


epezent:
Hi @sayan1an. The getter signatures are non-const to match the style of ImGui::PlotLines, etc. I imagine this decision was made because of the off-chance that a user-defined getter needs to modify the data as a part of the getting process. Therefore, I will not be merging this change.

I will add the MSVC check manually.

PR #15, opened 2020-05-09
PlotDigital multi level

ozlb:
Plot digital channels as float (only positive values)
Untitled


epezent:
This looks awesome. I'll give it a review and merge soon.

epezent:
Sorry but I can't merge this with the QueryX code in there. Can you factor it out and create a separate issue for QueryX? I have some thoughts about how we might allow for QueryX the way you want it.

ozlb:
ok. i will fix now the ImPlotFlags_NoChild issue and remove QueryX

PR #17, opened 2020-05-09
Update deprecated headers

sayan1an:
Headers now conform to C++11 standards.

### DIFF
--- a/implot_demo.cpp
+++ b/implot_demo.cpp
@@ -27,8 +27,8 @@
 #endif
 
 #include <implot.h>
-#include <math.h>
-#include <stdio.h>
+#include <cmath>
+#include <cstdio>
 
 namespace {
 


epezent:
Again, this was done to maintain conformance with `imgui.h` and `imgui_internal.h`. ImGui explicitly avoids C++11 features: 

> Dear ImGui takes advantage of a few C++ languages features for convenience but nothing anywhere Boost insanity/quagmire. Dear ImGui does NOT require C++11 so it can be used with most old C++ compilers. Dear ImGui doesn't use any C++ header file. Language-wise, function overloading and default parameters are used to make the API easier to use and code more terse. Doing so I believe the API is sitting on a sweet spot and giving up on those features would make the API more cumbersome. Other features such as namespace, constructors and templates (in the case of the ImVector<> class) are also relied on as a convenience.

PR #22, opened 2020-05-10
Remove uses of the C++11 auto feature

jpieper:


epezent:
This just made me realize we probably need to remove usage of range based for loops as well.

jpieper:
Yep.  Same with the few inline initializers that were present.  I can do both, but that will conflict more with my in progress work, so you mind if I wait a bit?

epezent:
Absolutely. Thank you!

PR #23, opened 2020-05-10
Identify NANs with isnan()

r0mai:
In C/C++ NAN == NAN is false, so isnan() must be used to identify NAN values.


epezent:
Thanks! I learned something new today.

PR #25, opened 2020-05-11
Support multiple Y axes simultaneously

jpieper:
This allows up to 3 Y axes to be used, one on the left side, and up to
2 on the right.  To use multiple axes, one of the `ImPlotFlags_Y2Axis`
or `ImPlotFlags_Y3Axis` must be passed to `BeginPlot`.  Then, before
each Plot call, SetPlotYAxis may be used to set that plot on an
alternate axis.

An extra line and ticks are rendered off to the right if all 3 axes
are configured (although those ticks are always only 5 long,
regardless of major status).  Each of the axes, when present, has a
(possibly invisible if ticks and labels are disabled) hit target that
can be used to scroll and scale that axis independently.

Some other changes were required to make this happen:
 * Queries are now only in pixel coordinates
 * ImPlotRange has been renamed to ImPlotBounds (and correspondingly
   GetPlotRange is now GetPlotBounds)
 * All APIs which allow querying values in plot coordinates now have
   an optional "y_axis" argument, which can be -1 to use the currently
   selected axis, or non-negative to pick a specific axes.
 * BeginPlot used to lock all flags on the first call.  Now it only
   locks the axis flags, and not the primary plot flags.
 * The mouseover text now renders the 2nd and 3rd axis in parenthesis.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -65,21 +65,33 @@ ImPlotStyle::ImPlotStyle() {
     Colors[ImPlotCol_PlotBorder]    = IM_COL_AUTO;
     Colors[ImPlotCol_XAxis]         = IM_COL_AUTO;
     Colors[ImPlotCol_YAxis]         = IM_COL_AUTO;
+    Colors[ImPlotCol_Y2Axis]        = IM_COL_AUTO;
+    Colors[ImPlotCol_Y3Axis]        = IM_COL_AUTO;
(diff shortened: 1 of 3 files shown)


jpieper:
20200510-multi-y-axis-2

epezent:
Found a few issues. Merged into `dev` branch for cleanup.

PR #33, opened 2020-05-13
call getter, transformer and cull test once per point instead of twice

sergeyn:
On my system this bumps FPS from 360 to 410 using stress test plots in demo function

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -2081,34 +2081,43 @@ inline void RenderLineAA(ImDrawList& DrawList, const ImVec2& p1, const ImVec2& p
 }
 
 template <typename Transformer, typename Getter>
-inline void RenderLines(Transformer transformer, ImDrawList& DrawList, Getter getter, int count, int offset, float line_weight, ImU32 col_line, bool cull) {
+inline void RenderLineStrip(Transformer transformer, ImDrawList& DrawList, Getter getter, int count, int offset, float line_weight, ImU32 col_line, bool cull) {
 // render line segments
+   offset %= count; 
+   if (offset < 0) offset += count; // shift negative offset to positive range
+   int i_start = offset + 1;
+   if (i_start >= count ) i_start -= count;
+   int i_end = offset + count;
+   if (i_end >= count) i_end -= count;
+   
     const int    segments  = count - 1;
-    int    i1 = offset;
-    ImVec2 p1, p2;
+    ImVec2 p1 = transformer(getter(offset));
+    bool test1 = !cull || gp.BB_Grid.Contains(p1);
     if (HasFlag(gp.CurrentPlot->Flags, ImPlotFlags_AntiAliased)) {
-        for (int s = 0; s < segments; ++s) {
-            const int i2 = (i1 + 1) % count;
-            p1 = transformer(getter(i1));
-            p2 = transformer(getter(i2));
-            i1 = i2;
-            if (!cull || gp.BB_Grid.Contains(p1) || gp.BB_Grid.Contains(p2))
+        for (int i1 = i_start; i1 != i_end; i1 = i1 + 1 < count ? i1 + 1 : i1 + 1 - count)
+        {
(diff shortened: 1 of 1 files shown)


epezent:
Solid optimization. I should have thought of that! I also like the negative offset. I added an example to the demo to demonstrate this. We should extend this to the other functions which take an offset value.

sergeyn:
Honestly speaking the whole offset thing does not belong there. The offset,
(and axis modifiers  like log scale) should move to  'getter' class.
RenderLine (now RenderLineStrip) function is the one taking 90% of the cpu
cost and it should be as straightforward as possible.
I'll propose a PR moving offset and axis scalers (read - your
'transformer') under the common 'getter' interface a bit later.

ozlb:
Maybe offtopic, but that there is still an issue with "extreme zoom" and this happens when a single segment has points out of plot area. Any idea?
Peek 2020-05-13 17-44

ozlb:
> > Maybe offtopic, but that there is still an issue with "extreme zoom" and this happens when a single segment has points out of plot area. Any idea?
> 
> That's exactly this problem. If you make sure 'culling' parameter is always false, the problem goes away. This would also makes markers disappear as soon as point goes out of culling rectangle. The whole culling thing needs to be revisited (and if it's needed in the first place).

Raw idea: 
- detect during for cycle the segment that is "crossing" (when p1.x is < min and p2.x is > max) and "save" it 
- after for cycle, if any plot has been done, plot the saved one

epezent:
> The whole culling thing needs to be revisited (and if it's needed in the first place).

I think if the culling process is kept cheap, then it should be done. What I don't want to do is introduce a complicated culling method that's expensive just to satisfy extreme zoom cases (e.g. a line intersection test between every segment and the plot boundaries) . The trade off doesn't seem worth it. 

I'm curious how many FPS can be regained just be eliminating the current culling process. Maybe it really isn't worth it. I'm trying to remember why we even introduced it.

sergeyn:
Doesn't ImGui do the culling ?
Simplest I can think of is a bbox test.

ozlb:
> > The whole culling thing needs to be revisited (and if it's needed in the first place).
> 
> I think if the culling process is kept cheap, then it should be done. What I don't want to do is introduce a complicated culling method that's expensive just to satisfy extreme zoom cases (e.g. a line intersection test between every segment and the plot boundaries) . The trade off doesn't seem worth it.
> 
> I'm curious how many FPS can be regained just be eliminating the current culling process. Maybe it really isn't worth it. I'm trying to remember why we even introduced it.

Are you kidding?! Because without, you will have "infinite" vtx plotted every time for nothing

Peek 2020-05-13 18-18

ozlb:
> I'm curious how many FPS can be regained just be eliminating the current culling process. Maybe it really isn't worth it. I'm trying to remember why we even introduced it.

https://github.com/mahilab/mahi-gui/issues/8#issuecomment-612547534

ozlb:
If you disable cull data..

Screenshot from 2020-05-13 18-26-07

epezent:
Haha! Fair enough. Like I said, I'm open to other methods as long as it's simple and cheap.

sergeyn:
I was under the impression that ImGui does per-primitive culling. If that
doesn't happen, then it has to be done somewhere, either on gpu or on cpu
side.

ozlb:
> I think if the culling process is kept cheap, then it should be done. What I don't want to do is introduce a complicated culling method that's expensive just to satisfy extreme zoom cases (e.g. a line intersection test between every segment and the plot boundaries) . The trade off doesn't seem worth it.

There is another issue with extreme zoom. There is a point where delta sum on float is too small due to IEE754 resolution and delta by wheel results stacked (@jpieper also for y axis).

sergeyn:
I've submitted PR#35 to address culling issue. For the actual extreme zoom I think double precision scaling would fix it.

epezent:
Yea, I've been considering switching to double under the hood. I'll take a look at 35

PR #37, opened 2020-05-19
Fix implicit conversion warnings (double to float) on MSVC /W4

PazerOP:
The `log10`, `sin`, and `cos` functions in the global namespace are the C versions from `<math.h>`. Since C does not support overloading, these functions take in a double and return a double https://en.cppreference.com/w/c/numeric/math/log10. This triggers warnings with an appropriately high warning level, which I use.

The same functions in the std:: namespace (in `<cmath>`) have overloads that allow them to take in a float and return a float https://en.cppreference.com/w/cpp/numeric/math/log10.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -58,6 +58,8 @@ You can read releases logs https://github.com/epezent/implot/releases for more d
 #include "implot.h"
 #include "imgui_internal.h"
 
+#include <cmath>
+
 #ifdef _MSC_VER
 #define sprintf sprintf_s
 #endif
@@ -244,9 +246,9 @@ struct ImTick {
     double PlotPos;
     float  PixelPos;
     ImVec2 Size;
-    int    TextOffset;    
+    int    TextOffset;
     bool   Major;
-    bool   RenderLabel;    
+    bool   RenderLabel;
 };
 
 struct ImPlotItem {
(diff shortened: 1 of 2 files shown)


sergeyn:
Hi,

It is enough to just include <cmath>. You don't need to add std:: everywhere since cmath pulls in overloads of these functions for floats and warnings go away.

Regards.

PazerOP:
@sergeyn I could be totally wrong on this, but aren't the overloaded versions only supposed to exist in the `std::` namespace? Since the C versions occupy the global namespace and cannot be overloaded.

epezent:
@PazerOP , I believe you are correct. Soon, I will be pushing internal changes that should resolve these issues. Apologies for the slow response to this issue.

epezent:
Hi, this should have been resolved by https://github.com/epezent/implot/commit/aa23d95e21cb261e933995891049ea8c8f213cd1

I checked this with all warnings on MSVC, GCC, and Clang, but do let me know if there are still issues.

sergeyn:
@PazerOP  You can overload extern"C" functions in global namespace as far as your overloads are not extern "C".  (at least on microsoft compiler, I didn't check other compilers)

PazerOP:
@sergeyn I am aware that you can do that, I was questioning whether the C++ standard defines these overloads in the global namespace as well as in the `std` namespace.

Issue #34, opened 2020-05-13
Addition of support to plot time along the x-axis


Aniket025:
I was wondering if there is something to support time in dd/mm/yy and hours:minutes: seconds on the x-axis vs some data on the y-axis.

epezent:
Other people have asked for time-formatted x-axes, so I think it's probably a feature that could be useful and something we can start looking into. I see two paths forward:

1) Expose a public API that allows the user to define their on axes label formatting. I don't know how it would look just yet, but this would be the most flexible option. 

2) Support time format directly (in addition to normal and scientific formats). If we go this route, then other questions arise: 
  - do we treat the user's `float* xs` input as seconds? what if their data is in minutes or hours?
   - are the labels fixed to a certain format (e.g hh::mm::ss.ms) or dynamic based on the range (e.g. switch from hour to min to sec as the user zooms in)
   
Either way, we will need a more sophisticated grid division/subdivision approach that only subdivides if labels do no overlap, or hides labels when they do overlap.  This should be our first focus.

Aniket025:
I think there can be possible solutions by going with the second option. 
People who want to plot will mostly have it as a timestamp which can be converted to whatever the user wishes to have. Also, plotting every individual second on the x-axis won't be a very good option. 
We can have a specific label as HH: MM initially and as the user zooms we can add HH:MM:: SS and then HH:MM::SS: MS to it. 
I am sharing a snapshot of an application that does the same and we can make something similar along the x-axis. 
Please see the images below : 
![image](https://user-images.githubusercontent.com/25415857/81887963-f7626600-95bd-11ea-8ce0-6f4dcddb1261.png)
Here they have in HH:MM
Then as we zoom 
![image](https://user-images.githubusercontent.com/25415857/81888034-1cef6f80-95be-11ea-9632-9e281180be18.png)

We can follow a similar approach which shall solve major issues for most people.

Aniket025:
Also, we might keep the gridlines constant for a frame for better visibility and divide a given time into a specific number of slots and mark their labels to help make it dynamic.

Prinkesh:
Hi @epezent  . Hi have added support for time axis with formatting based on current data range .
Max granularity that one can see is of milliseconds .


![](https://i.gyazo.com/6137bddfd3dffbf711e5446d7f327df5.gif)

[...]

PR #38, opened 2020-05-20
Feature/timestamp support

Prinkesh:
* Added Support for TImeStamp Support for XAxis . 
* TIme Input should be in microseconds 
* Fixed API bugs for double api

### DIFF
--- a/TimeFormatter.hpp
+++ b/TimeFormatter.hpp
@@ -1,0 +1,117 @@
+#ifndef _INC_TIMEFORMATTER_HPP_
+#define _INC_TIMEFORMATTER_HPP_
+#include <iostream>
+#include <ctime>
+#include <sstream>
+#include <iomanip>
+
+class TimeFormatter {
+    public:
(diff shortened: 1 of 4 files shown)


Prinkesh:
Related Issue #34

epezent:
Awesome! I'll try to get to this in the next day or so. Can you post some GIFs in the meantime.

epezent:
@Prinkesh , I am unable to try this code due to a build error (MSVC):  `Cannot open include file: 'sys/time.h': No such file or directory`

Otherwise, my feedback is:
1) Move `TimeFormatter` into `implot.cpp`. We don't want an extra header for this.
2) Remove all C++11 headers and containers (`std::string` is problematic in particular)
3) Find a portable alternative for `sys/time.h`

I still think this is a necessary feature. Please post GIFs so we can see how this works. Thanks!

epezent:
Sorry for the close. This was an unintended side-effect of deleting the double branch. Please resubmit.

PR #44, opened 2020-05-27
Fix implicit double to float conversions.

exjam:
You do not include <cmath> in implot.cpp and instead rely on the include of math.h through the imgui_internal.h header - this C version of the header not have the float overloads leading to double to float conversion warnings.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -244,9 +244,9 @@ struct ImTick {
     double PlotPos;
     float  PixelPos;
     ImVec2 Size;
-    int    TextOffset;    
+    int    TextOffset;
     bool   Major;
-    bool   RenderLabel;    
+    bool   RenderLabel;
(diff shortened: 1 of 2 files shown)


exjam:
Oh crap I just saw #37 is already open for this ...

exjam:
Well I guess feel free to merge either one depending on the fix you prefer, whether that is include <cmath> or not.

epezent:
Thanks. I'm currently working on some internal changes that will probably obviate this. I'll leave it open for now.

epezent:
Hi, this should have been resolved by https://github.com/epezent/implot/commit/aa23d95e21cb261e933995891049ea8c8f213cd1

I checked this with all warnings on MSVC, GCC, and Clang, but do let me know if there are still issues.

PR #55, opened 2020-06-07
Feature/timestamp support rebase

### LINKED ISSUE #34: Addition of support to plot time along the x-axis

Prinkesh:
Closes #34


epezent:
1) Can you resolve the current conflicts?
2) Is there any reason we can't use `seconds` instead of `microseconds`?

Prinkesh:
> 2\. Is there any reason we can't use `seconds` instead of `microseconds`?

We can use seconds . But then we will then have to let the user define min granularity of data . Probably by adding flags or functions for setting granularity . Now in that case we will have to add support for higher granularity like days as well if we are adding support for seconds . 

Since microseconds is the max granularity we can support without adding support for other data types or offsetting the data so  i went with that .

epezent:
@Prinkesh, this is off to a solid start! I have a few observations and suggestions:

- Crashes when zooming out beyond MM/DD
- Subdivision labels for the most part a nicely rounded, but there are some cases when unintuitive values are chosen:

image

image

- The bottom left label isn't pixel aligned with the right
- I would suggest truncating the bottom left label based on what granularity is displayed on the axis labels. E.g., here the bottom left label could just be `2020/05/20 15:56, and then the second information is given in the axis labels. 

image

- You might consider matching the format of the bottom right label (mouse hover) to the format displayed on the axis (i.e. just 30.1, not the full 2020/05/20 12:56:30.100000.
- Trailing zeros should be eliminated where possible using `%g` format.

Prinkesh:
@epezent Hi I have fixed the issues and added better tick units determination for major and minor ticks .

Prinkesh:
@epezent Hi I have fixed the issues and added better tick units determination for major and minor ticks .

epezent:
Awesome, I'll take a look!

epezent:
@Prinkesh, this is looking really good. I found a few more peculiar things:

* When transitioning from days to hours, the day is not displayed in the bottom left label:

hour_day

* 45 is being chosen as an interval for minutes (and presumably seconds as well). Since this isn't a factor of 60, it causes axes to suddenly pop in/out. I think only intervals of 15,30, and 60 should be allowed.



* Days don't display a leading zero:

image

* [Day hour] is using an interval of 15 for hours, which is not a factor of 24 (I think the most logical set of intervals is 1, 2, (maybe 4), 6, 12)

image

* When years are displayed on the x-axis, I noticed that the bottom left label would randomly display a very large number at times when zooming. I wasn't able to capture this with a GIF because it happens so quickly and doesn't seem to stick. 

Thanks for your continued work on this. I know it's probably been very frustrating to program.

Prinkesh:
Have changed the step sizes as suggested and fixed the padding issue in dates 

> When years are displayed on the x-axis, I noticed that the bottom left label would randomly display a very large number at times when zooming. I wasn't able to capture this with a GIF because it happens so quickly and doesn't seem to stick.

This is because the time range goes below `1st January 1970` minimum supported time for formatting . Have changed to show no values if this happens and on next iteration the range is contrained to minimum supported time .

Prinkesh:
@epezent Did you look at changes ?

epezent:
I did. It looks good! I don't see anything strange anymore. I do want to review the code before merging, so I may be a day or two longer. Thanks for being patient.

epezent:
I haven't forgotten about this. Just very busy with more pressing matters. Hope to merge soon.

PR #58, opened 2020-06-07
Vertical and Horizontal error bars

### LINKED ISSUE #57: Addition of horizontal error bars

ChiefSeaBiscuit:
Closes #57

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -2920,29 +2920,6 @@ void PlotErrorBars(const char* label_id, const double* xs, const double* ys, con
 // PLOT ERROR BARS H
 //-----------------------------------------------------------------------------
 
-struct ImPlotPointErrorH {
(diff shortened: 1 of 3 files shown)


ChiefSeaBiscuit:
An example of the additional demo can be seen in #57

epezent:
@ChiefSeaBiscuit, looks good. One suggestion: I don't think it's necessary for the horizontal version to also do the vertical rendering. If the user wants vertical and horizontal bars, they just make two function calls. Can you make PlotErrorBarsH only do horizontal? As such, you won't need `ImPlotPointErrorH `, as you can reuse `ImPlotPointError`.

ChiefSeaBiscuit:
@epezent Ahh yeah true, that makes more sense now you say it! I'll try and get those changes in tonight or so!

ChiefSeaBiscuit:
Added the changes, I've also altered the demo for it slightly just to highlight that positive and negative errors can be set individually as the developer might not be aware of this feature, from the original demo alone without looking at the function decs

image

PR #59, opened 2020-06-08
PlotDigitalEx fix NAN values and fix circular buffer rollover (issue #60)

ozlb:
PlotDigitalEx fix NAN handling and behaves as other plot functions (i use NAN for NULL samples)


epezent:
Using NAN for null samples or samples to skip is useful for other plot types as well. How can we handle this generically for all plots? Presumably it should go somewhere in Getters.

ozlb:
Actually PlotLine is already handling somehow NAN.
In the meantime I fixed also Issue https://github.com/epezent/implot/issues/60

ozlb:
If you want to handle in Getter functions probably it should be modified from "for cycle up to count" to "while !endOfBuffer do getNext"

PR #68, opened 2020-06-22
Only display non-empty legends

jpieper:
Does this seem reasonable?  I wanted some plot items to have no legend, and just prefixing them with `##` otherwise leaves a blank colored line in the legend bar.


epezent:
Seems reasonable to me. Not much to gained from an unlabeled legend entry. I'll check this for any unintended side-effects later this afternoon and merge.

PR #71, opened 2020-06-24
Fix divison by zero when plotting 0 data lenght.

JoelLinn:


epezent:
@JoelLinn , can you verify that this PR doesn't cause an significant slow downs. You can try testing the benchmark in the demo with v-sync turned off. Let me know if there's an FPS hit.

JoelLinn:
I removed the check from `OffsetAndStride` since it afaik is not called by any plot function if the count is 0 and it caused a 3% fps drop in the benchmark on Linux.
The added check now only happens one time when the getter is constructed per plot* call. With this the fps drop in the benchmark is immeasurable for me.

PR #85, opened 2020-08-07
Fit Data function PR

### LINKED ISSUE #83: Addition of a Fit Data function
I was thinking of adding a function which performs the exact same operation as a double click on a graph to fit the data but as a function call, thoughts? Should be a simple addition (If not already in and I've somehow missed it)

Could also include a new enum to specific which axes you want to fit specifically
```cpp
enum ImPlotAxis_{
X_AXIS,
Y1_AXIS,
Y2_AXIS,
Y3_AXIS,
ALL
}
```

Simple function declaration with default to do all
```cpp
ImPlot::FitData(ImPlotAxis_ axis = ImPlotAxis_::ALL);
```

ChiefyChief23:
Closes #83


epezent:
@ChiefyChief23 , in your demo code, it's a bit weird to see `ImGui::Button` inside of the ImPlot `if` statement. Does this function still work as expected if it is called outside of the `if` statement, before `BeginPlot`?

Also, can the function implementation be simplified, e.g, `gp.FitX = fit_x_axis;` etc. or will this have some unintended side effect I'm not seeing.

And finally, can I propose a rename to `FitAxes`?

ChiefyChief23:
The button is just to individual choose when to fit what axes and just to show the user how it would work. I can remove that out though and refactor it slightly. I'll test it out in different places to confirm what placement is works and not as well.

I'll get the other two changes in too.

ChiefyChief23:
The code will only fit the data to an axis if the function call is within a BeginPlot and before the plot function calls are made. I've updated the demo to show this and updated the function declaration comment. Thoughts?

epezent:
Hmm, I see. I think this approach would be prone to misuse. We might want this to be a function that is called before `BeginPlot` e.g. `FitNextAxes` so that there's no confusion about the order of calls. This matches the style of `SetNextPlotLimits` etc. as well.

epezent:
I added `FitNextPlotAxes` to the API. See https://github.com/epezent/implot/commit/7da6ed69f3458db6af85c3215f4e06676bcf2a4f

PR #88, opened 2020-08-13
Fix overflow when using 32-bit ImDrawIdx

bear24rw:
For backends that don't support `VtxOffset` (the imgui emscripten example) we are forced to use a 32bit index buffer instead. This patch fixes an issue from #41 where the calculation of the maximum index would overflow. The benchmark plot now renders correctly in the emscripten example when compiling with `-DImDrawIdx="unsigned int"`.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -2286,15 +2286,16 @@ struct TransformerLogLog {
 /// Renders primitive shapes in bulk as efficiently as possible.
 template <typename Renderer>
 inline void RenderPrimitives(Renderer renderer, ImDrawList& DrawList) {
-    int prims = renderer.Prims;
-    int prims_culled = 0;
-    int idx = 0;
+    unsigned int prims = renderer.Prims;
+    unsigned int prims_culled = 0;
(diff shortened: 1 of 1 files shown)


epezent:
Thanks for the fix! Can you make `max_idx` either static or a compile time constant so it's not reevaluated each time.

bear24rw:
Good idea, I just force pushed that change.

epezent:
@bear24rw , I don't know if you noticed, but I ended up changing your PR slightly. I replaced your `ImPow` calculation with some template code:

```cpp
template <typename T>
struct MaxIdx { static const unsigned int Value; };
template <> const unsigned int MaxIdx<unsigned short>::Value = 65535;
template <> const unsigned int MaxIdx<unsigned int>::Value   = 4294967295;
```

```cpp
...
unsigned int cnt = ImMin(prims, (MaxIdx<ImDrawIdx>::Value - DrawList._VtxCurrentIdx) / Renderer::VtxConsumed);
...
```

Does this still work as expected for you?

PR #90, opened 2020-08-16
Fix yaxis setting updates not working

PeterJohnson:
The new values were being stored in PreviousFlags instead of Flags.

I don't understand the PreviousFlags logic in general, is the idea to have PreviousFlags be the flags before any double-right-click popup menu modifications?

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -852,11 +852,11 @@ bool BeginPlot(const char* title, const char* x_label, const char* y_label, cons
         if (flags != plot.PreviousFlags)
             plot.Flags = flags;
         if (y_flags != plot.YAxis[0].PreviousFlags)
-            plot.YAxis[0].PreviousFlags = y_flags;
+            plot.YAxis[0].Flags = y_flags;
         if (y2_flags != plot.YAxis[1].PreviousFlags)
-            plot.YAxis[1].PreviousFlags = y2_flags;
+            plot.YAxis[1].Flags = y2_flags;
         if (y3_flags != plot.YAxis[2].PreviousFlags)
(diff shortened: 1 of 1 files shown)


epezent:
Thanks, this would have resulted in a bug for those axes. Since the state of flags can be changed via the context menu, the plot's flags are cached into `plot.axis.Flags` during the first call to `BeginPlot`. It is this copy that is subsequently modified by the context menu. However, users might decide to change the flags themselves in later calls to `BeginPlot`, so we have to do some book keeping to detect when this happens. Hence, `PreviousFlags`.

PeterJohnson:
As a follow-up to this, would it be better to have the axis and plot flags as style variables (e.g. in ImPlotStyle and via PushStyleVar) instead of parameters to BeginPlot()?  It feels like that would be more extensible, and right now it's not possible to read back many of the settings that can be changed via the context menu.

epezent:
I've considered it, but usually ImGui functions expect the flags as an argument to the widget function. 

Is reading back the altered flags state something that would be useful? I'm trying to imagine an actual use case...

PeterJohnson:
My application is a GUI that runs within the user’s application, and is frequently restarted by the user (code/debug cycle), so I want to persist all the settings so the user feels like they’re coming back into the GUI where they left off, and not having to tweak all the settings again.  Right now I’m putting settings on a separate window but it would be nice if the context menu setting changes were integrated to be persistent as well (where appropriate).

epezent:
Gotcha. I'm currently refactoring some of the code, and exposing more of ImPlot's guts in a new file `implot_internal.h` to enable more advanced use cases like yours. With this, you should be able to get all the information you need. Expect this relatively soon.

PR #96, opened 2020-08-19
Draw query rect only when query enabled.

jaswantp:
The query rectangle is drawn regardless of whether ImPlotFlags_Query is included within ImPlotFlags.
This is quite unintuitive when I do a query and later disable query mode. It would be cleaner if the query rectangle disappears altogether!

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -1345,20 +1345,23 @@ void EndPlot() {
         }
     }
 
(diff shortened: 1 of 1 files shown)


epezent:
An oversight on my part. Thanks!

jaswantp:
Nevertheless, This is an excellent tool. Thumbs up to the simple design. :) Thanks.

PR #97, opened 2020-08-19
Exposed RegisterItem and ImPlotList struct

hoffstadt:
The purpose of this pull request is to expose **RegisterItem** and the **ImplotList** struct to allow users to added custom render items to the legend. For example, I implemented a general polygon fill custom rendering and wanted it to appear in the legend and be able to hide and show it. Here you can see it working:

* _Sin_ is the custom render item.

Capture

Here is the code in use:
```cpp
ImPlot::PushPlotClipRect();
auto item = ImPlot::RegisterItem("Sin");
if(item->Show)
    DrawPolygon(ImGui::GetWindowDrawList(), m_xs, m_ys);
ImPlot::PopPlotClipRect();
```

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -195,6 +195,18 @@ struct ImPlotInputMap {
     ImPlotInputMap();
 };
 
+// State information for Plot items
+struct ImPlotItem {
+    ImPlotItem();
+    ~ImPlotItem();
+    bool Show;
+    bool SeenThisFrame;
+    bool Highlight;
+    ImVec4 Color;
+    int NameOffset;
+    ImGuiID ID;
+};
+
 //-----------------------------------------------------------------------------
 // ImPlot End-User API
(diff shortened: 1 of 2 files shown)


epezent:
Any particular reason you needed this to be in `implot.h`? Unless I'm missing something, it should still be possible to do this by including `implot_internal.h`. I'd rather keep the struct there for now. Maybe in the future we will pull some of this into the public API.

hoffstadt:
I see. At the time when I actually implemented this, it was in v0.4 which did not yet have implot_internal.h. I just made the change to implot_internal.h without realizing I could just include that now. Thanks for the quick response.

epezent:
Yep! Also, note the new demo section `Custom Plotters and Tooltips`. You may find something interesting there.

Issue #89, opened 2020-08-15
Feature request: drag from legend


PeterJohnson:
My application has multiple visible plots.  It’s really easy to drag from a data source to a particular plot, but it would also be a nice feature to be able to drag a series from the legend of one plot to another plot.  API wise I'm thinking something like (between BeginPlot and EndPlot), mimicing ImGui::BeginDragDropSource():
```
if (ImPlot::BeginLegendDragDropSource("id")) {
  ImGui::SetDragDropPayload(...);
  // text etc for DND tooltip
  ImPlot::EndLegendDragDropSource();
}
```

I think a temporary workaround would be to use the legend hovered feature and detect drags, but that feels very manual and error-prone.

epezent:
Interesting idea! If you can submit a working prototype that wraps the current API (e.g. using legend hovered API as you suggested) and/or minimal changes to the inner workings, I'd be happy to consider it.

ozlb:
> Interesting idea! If you can submit a working prototype that wraps the current API (e.g. using legend hovered API as you suggested) and/or minimal changes to the inner workings, I'd be happy to consider it.

@epezent 
I think that is missing also a mechanism (function SetNext___ or something) in order to temporary disable axis drag&drop without locking. This is a must if you want to drag&drop something that has been rendered after BeginPlot. What do you think, what is your idea about?

PeterJohnson:
I have this working and will PR it soon.  It was pretty straightforward to do based on the ImGui::BeginDragDropSource() implementation.  It's somewhat magical.  It's not necessary to disable axis drag and drop for it to work.

epezent:
Excellent! Can't wait to see the PRs

PeterJohnson:
Closed by #106.

PR #98, opened 2020-08-19
SetNextPlotDragDisable

ozlb:
Set drag operations disable for the next plot. Call right before BeginPlot(). 
https://github.com/epezent/implot/issues/89#issuecomment-675132076


epezent:
I understand what this does, but I don't understand why it's needed. Alternatively, the axis flags could be temporarily modified to include `ImPlotAxisFlags_LockMin` and `ImPlotAxisFlags_LockMax`. What particular problem are you trying to solve?

ozlb:
Assume that you want to customize with "widget" a plot
Assume that you want to interact with that "widget" and drag it
a) you need to be after BeginPlot in order to get DrawList and to render it
b) drag of axis and query is inside BeginPlot so it's BEFORE the widget code
c) II need to lock axis but also potentially all other begin of drag operations, as the one mentioned from label or whatever will be in future

epezent:
Ok, that makes sense. Let me think on this and make sure there isn't a better solution. I can see potential headaches with the approach you described. Needing widget logic/input detection before `BeginPlot`, then `SetNextPlotDragDisable` conditioned on that logic, and then widget rendering after `BeginPlot` seems a bit burdensome. Perhaps it's time to properly implement custom a widget/tooltip system. Maybe via callbacks?

ozlb:
I’m with you that it’s not perfect. Callback could be a solution, you have a better vision of actual status and direction for future.

Issue #94, opened 2020-08-17
Gallery: Post Your Screenshots / Code Here (PART 1)


epezent:
Post your screenshots or custom code here and tell us about your project!

epezent:
This is a tool I wrote for [Tasbi](http://evanpezent.com/hardware/tasbi/), a haptic wristband I built at Facebook Reality Lab and use for my PhD research. I'm using ImPlot to visualize and tune motor controllers, and calibrate sensor measurements. ImPlot works great with ImGui's docking branch!

![Tasbi](https://user-images.githubusercontent.com/29577475/77241246-5064dd80-6bbd-11ea-9a9f-28ab981c17de.png)

epezent:
**Custom candlestick plots and tool tips:**

![candletool](https://user-images.githubusercontent.com/29577475/90851160-7b548880-e339-11ea-90b5-27b993421005.gif)

**ImPlot + ImGui Tables API:**

![tables](https://user-images.githubusercontent.com/29577475/90851187-8c04fe80-e339-11ea-8267-e6bfdc105686.gif)

Both of these have been added to `implot_demo.cpp`

PeterJohnson:
Robot simulation GUI for FIRST robotics with multiple plots and drag and drop everything.


![drag and drop resized](https://user-images.githubusercontent.com/297126/91929143-9a84db80-ec92-11ea-80e1-3eea3ff8b890.gif)


https://github.com/wpilibsuite/allwpilib/pull/2653

epezent:
@PeterJohnson , I'm curious how you got the linked axes working and if it is bidirectional. This is a feature I've wanted to implement.

hoffstadt:
Some things DearPyGui users have made:

![fa08r61ae2i51](https://user-images.githubusercontent.com/39973752/92046481-3fadbb80-ed48-11ea-9b3b-9cc5ea6e3a4e.png)

![114vv9n3ask51](https://user-images.githubusercontent.com/39973752/92046084-c910be00-ed47-11ea-9a33-7a761eb7d356.jpg)

![91245835-0c4daa00-e714-11ea-8b3f-6ac01d6dbeb3](https://user-images.githubusercontent.com/39973752/92046379-ffe6d400-ed47-11ea-9a19-ab672fa9c26f.gif)
[...]

PR #106, opened 2020-09-01
Add BeginLegendDragDropSource

PeterJohnson:
This uses pieces of the ImGui BeginDragDropSource implementation to
provide a method to drag from a legend entry.

EndDragDropSource is just a wrapper around ImGui::EndDragDropSource.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -1864,6 +1864,75 @@ bool IsLegendEntryHovered(const char* label_id) {
         return false;
 }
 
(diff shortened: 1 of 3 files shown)


PeterJohnson:
Fixes #89.

epezent:
Made a few minor edits, but otherwise good PR and really cool feature! I can imagine several scenarios where this could be used. If you can, would be mind posting GIFs of your more advanced use case in https://github.com/epezent/implot/issues/94. I recall you have an inner-plot drag and drop system working? 

Thank you!!

PR #107, opened 2020-09-02
Make EndItem not inline

PeterJohnson:
On some compilers (e.g. clang on Mac), this results in a linker error as
it's inlined and can't be found from other source files (e.g. implot_demo).


epezent:
Thanks, had to make the change to BeginItem too so just pushed it myself.

PR #116, opened 2020-09-17
Handle possibly overlapping widgets added via ImGui::SetItemAllowOverlap()

bear24rw:
For my application I needed a horizontal reference line that could be adjusted by dragging it with the mouse. To do this I added an invisible button and detect when that button is being dragged. In order for this to work I needed to modify ImPlot to tell `ImGui::ItemAdd` what our ID is, and then check for if `HoveredIdPreviousFrame` matches our ID. If it does not match that indicates that the mouse is actually hovering a widget added on top of us. I'm not sure why `ImGui::ItemHoverable` does not check this by default?

I've included a demo of a horizontal line movable via click and drag.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -1273,6 +1273,7 @@ bool BeginPlot(const char* title, const char* x_label, const char* y_label, cons
     gp.Hov_Frame = ImGui::ItemHoverable(gp.BB_Frame, ID);
     if (G.HoveredIdPreviousFrame != 0 && G.HoveredIdPreviousFrame != ID)
         gp.Hov_Frame = false;
+    ImGui::SetItemAllowOverlap();
     ImGui::RenderFrame(gp.BB_Frame.Min, gp.BB_Frame.Max, GetStyleColorU32(ImPlotCol_FrameBg), true, Style.FrameRounding);
 
     // canvas bb
@@ -1722,7 +1723,6 @@ bool BeginPlot(const char* title, const char* x_label, const char* y_label, cons
         }
     }
     ImGui::PopClipRect();
-
     // push plot ID into stack
     ImGui::PushID(ID);
     return true;
@@ -2483,9 +2483,9 @@ ImPlotLimits GetPlotQuery(int y_axis_in) {
     return result;
 }
 
-bool HorizontalGuide(const char* id, double* value, const ImVec4& col, float thickness) {
+bool GrabLineH(const char* id, double* value, const ImVec4& col, float thickness) {
     ImPlotContext& gp = *GImPlot;
-    IM_ASSERT_USER_ERROR(gp.CurrentPlot != NULL, "HorizontalGuide() needs to be called between BeginPlot() and EndPlot()!");
+    IM_ASSERT_USER_ERROR(gp.CurrentPlot != NULL, "GrabLineH() needs to be called between BeginPlot() and EndPlot()!");
     const float grab_size = ImMax(5.0f, thickness);
     float xl = gp.BB_Plot.Min.x;
     float xr = gp.BB_Plot.Max.x;
@@ -2493,7 +2493,6 @@ bool HorizontalGuide(const char* id, double* value, const ImVec4& col, float thi
     const bool outside = y < (gp.BB_Plot.Min.y - grab_size / 2) || y > (gp.BB_Plot.Max.y + grab_size / 2);
     if (outside)
         return false;
-
     float len = gp.Style.MajorTickLen.y;
     ImVec4 color = IsColorAuto(col) ? ImGui::GetStyleColorVec4(ImGuiCol_Text) : col;
     ImU32 col32 = ImGui::ColorConvertFloat4ToU32(color);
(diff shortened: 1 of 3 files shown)


epezent:
Awesome! I had no idea it was possible to capture drag this way. This is an issue @ozlb has been talking about for a long time, also for dragable guides/line (#113, #98). This PR pretty much solves the main issue we were having -- disabling plot input when a user tool or widget is engaged.

The line/guide tool is so useful and requested, I may clean up and incorporate your demo code into the public API, e.g. `ImPlot::HorizontalGuide(const char* id, double* value)`. 

Stay tuned...

ozlb:
Actually yes, it looks like we do have identical request (in my case 2 vertical lines)!

epezent:
I just added the following:

```cpp
// Shows a draggable horizontal guide line.
IMPLOT_API bool HorizontalGuide(const char* id, double* value, const ImVec4& col = IMPLOT_AUTO_COL, float thickness = 2);
// Shows a draggable vertical guide line.
IMPLOT_API bool VerticalGuide(const char* id, double* value, const ImVec4& col = IMPLOT_AUTO_COL, float thickness = 2);
```

Demo:

```cpp
if (ImPlot::BeginPlot("##guides",0,0,ImVec2(-1,0),ImPlotFlags_YAxis2)) {            
    ImPlot::VerticalGuide("v1",&v1, ImVec4(0,1,0,1));
    ImPlot::VerticalGuide("v2",&v2, ImVec4(0,1,0,1));
    ImPlot::HorizontalGuide("h1",&h1, ImVec4(1,0,0,1));
    ImPlot::HorizontalGuide("h2",&h2, ImVec4(1,0,0,1));
    ImPlot::SetPlotYAxis(1);
    ImPlot::HorizontalGuide("h3",&h3, ImVec4(1,1,0,1));
    ImPlot::EndPlot();
}
```

guides


@ozlb , can you recommended any additional functionality? e.g. I was thinking it might be useful to display a tick label for the current value.

bear24rw:
Thanks for the fast turn around getting a real api around it! Looks perfect for my use case. I think the tick label could easily be drawn by the user with PlotText()?

ozlb:
> @ozlb , can you recommended any additional functionality? e.g. I was thinking it might be useful to display a tick label for the current value.

If you mean something similar to what @Prinkesh show in https://user-images.githubusercontent.com/5650130/83980205-720a7100-a931-11ea-9589-7283f57f7d77.gif (red label in x axis values), I think it will be very nice and if before the value there will be also the char *id that identifies the gudie (i.e. v1=0.3,  v2=0.7), even better!

One question: imagine that i would like to find the "closest" x[] and so [y] of a PlotLine relative o a vertical guide, how you suggest to find it? As per candlestick demo, with "binary search" strategies?

epezent:
Added labels and a better demo. Thoughts?

guides

epezent:
> One question: imagine that i would like to find the "closest" x[] and so [y] of a PlotLine relative o a vertical guide, how you suggest to find it? As per candlestick demo, with "binary search" strategies?

Something like that. I'd probably use `std::lower_bound` or `std::upper_bound` in practice.

epezent:
We should also introduce an "anchor" tool -- essentially a point that is draggable in both dimensions. You could imagine using several of these as spline handles for timeline/animation editors, etc.

bear24rw:
Love the label and the extra thick tick at the end, makes it stand out over a regular line. Draggable anchors would definitely be useful! I also wonder if you use this feature to make the query rect easily adjustable via click and drag.

bear24rw:
Noticed a minor issue: when drawing the zoom box with right click + drag, when your mouse hovers over the guide lines it changes the cursor and shows the label.

epezent:
> Noticed a minor issue: when drawing the zoom box with right click + drag, when your mouse hovers over the guide lines it changes the cursor and shows the label.

Fixed

epezent:
Here's a prototype for anchors!

```cpp
// Shows a draggable anchor point. #col defaults to ImGuiCol_Text.
bool AnchorPoint(const char* id, double* x, double* y, const ImVec4& col = IMPLOT_AUTO_COL, float radius = 5);
```

anchor

ozlb:
> We should also introduce an "anchor" tool -- essentially a point that is draggable in both dimensions. You could imagine using several of these as spline handles for timeline/animation editors, etc.

you can use also to pin an event during data recording or to notify an event... and since they are responding to `(ImGui::IsItemHovered() || ImGui::IsItemActive())` there are infinite solutions (i.e. trap double click to remove or change color or whatever interaction).
You "open a Pandora's box"...

ozlb:
At the limit of the “off topic” comment.
With new API released after this PR I completely removed my “QueryX” implementation that i was maintaining during all this time for each new version of library. I’m very happy to use ImPlot in native way!

epezent:
I'm glad we finally found a way to support this feature for you @ozlb! I'd like to see if we can migrate regular Query to this style (i.e. decouple from Begin/EndPlot and make it a separate "tool" function).

PR #122, opened 2020-09-21
Performance impact of offset and stride / wrap around

JoelLinn:
### The problem
Just indexing the data (patched out ImPosMod() with just idx) gives a 30% fps boost in the benchmark (mahi-gui) compared to the current implementation honouring offset, stride. Before ImPosMod() [was patched into it](https://github.com/epezent/implot/commit/eff8b5be43e2625245058c0d7c38d1aa71d47dda), the custom getter function version was faster.
To my testing the penalty of using a getter function versus raw array pointers (all without ImPosMod and OffsetAndStride()) is surprisingly low (I suspect because there is already a lot of function calls for drawing each data point anyways).

### ~~The~~ A Solution
One way to improve this is to add overloads without offset and stride parameters (and special getters in implot_items.cpp) for all functions. This will lead to double the amount of template instantiations though. A leaner variant would be to add overloads just for the _G functions. This also gives about 30% more fps in the plots benchmark.

If you have a better idea how to solve this I'm happy to rebase the PR.

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -351,28 +351,33 @@ IMPLOT_API void EndPlot();
 // Plots a standard 2D line plot.
 template <typename T> IMPLOT_API void PlotLine(const char* label_id, const T* values, int count, double xscale=1, double x0=0, int offset=0, int stride=sizeof(T));
 template <typename T> IMPLOT_API void PlotLine(const char* label_id, const T* xs, const T* ys, int count, int offset=0, int stride=sizeof(T));
-                      IMPLOT_API void PlotLineG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, int offset=0);
+                      IMPLOT_API void PlotLineG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, int offset);
+                      IMPLOT_API void PlotLineG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count);
 
 // Plots a standard 2D scatter plot. Default marker is ImPlotMarker_Circle.
 template <typename T> IMPLOT_API  void PlotScatter(const char* label_id, const T* values, int count, double xscale=1, double x0=0, int offset=0, int stride=sizeof(T));
 template <typename T> IMPLOT_API  void PlotScatter(const char* label_id, const T* xs, const T* ys, int count, int offset=0, int stride=sizeof(T));
-                      IMPLOT_API  void PlotScatterG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, int offset=0);
+                      IMPLOT_API  void PlotScatterG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, int offset);
+                      IMPLOT_API  void PlotScatterG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count);
 
 // Plots a shaded (filled) region between two lines, or a line and a horizontal reference.
 template <typename T> IMPLOT_API void PlotShaded(const char* label_id, const T* values, int count, double y_ref=0, double xscale=1, double x0=0, int offset=0, int stride=sizeof(T));
 template <typename T> IMPLOT_API void PlotShaded(const char* label_id, const T* xs, const T* ys, int count, double y_ref=0, int offset=0, int stride=sizeof(T));
 template <typename T> IMPLOT_API void PlotShaded(const char* label_id, const T* xs, const T* ys1, const T* ys2, int count, int offset=0, int stride=sizeof(T));
-                      IMPLOT_API void PlotShadedG(const char* label_id, ImPlotPoint (*getter1)(void* data, int idx), void* data1, ImPlotPoint (*getter2)(void* data, int idx), void* data2, int count, int offset=0);
+                      IMPLOT_API void PlotShadedG(const char* label_id, ImPlotPoint (*getter1)(void* data, int idx), void* data1, ImPlotPoint (*getter2)(void* data, int idx), void* data2, int count, int offset);
+                      IMPLOT_API void PlotShadedG(const char* label_id, ImPlotPoint (*getter1)(void* data, int idx), void* data1, ImPlotPoint (*getter2)(void* data, int idx), void* data2, int count);
 
 // Plots a vertical bar graph. #width and #shift are in X units.
 template <typename T> IMPLOT_API void PlotBars(const char* label_id, const T* values, int count, double width=0.67, double shift=0, int offset=0, int stride=sizeof(T));
 template <typename T> IMPLOT_API void PlotBars(const char* label_id, const T* xs, const T* ys, int count, double width, int offset=0, int stride=sizeof(T));
-                      IMPLOT_API void PlotBarsG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, double width, int offset=0);
+                      IMPLOT_API void PlotBarsG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, double width, int offset);
+                      IMPLOT_API void PlotBarsG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, double width);
 
 // Plots a horizontal bar graph. #height and #shift are in Y units.
 template <typename T> IMPLOT_API void PlotBarsH(const char* label_id, const T* values, int count, double height=0.67, double shift=0, int offset=0, int stride=sizeof(T));
 template <typename T> IMPLOT_API void PlotBarsH(const char* label_id, const T* xs, const T* ys, int count, double height, int offset=0, int stride=sizeof(T));
-                      IMPLOT_API void PlotBarsHG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, double height,  int offset=0);
+                      IMPLOT_API void PlotBarsHG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, double height,  int offset);
+                      IMPLOT_API void PlotBarsHG(const char* label_id, ImPlotPoint (*getter)(void* data, int idx), void* data, int count, double height);
 
 // Plots vertical error bar. The label_id should be the same as the label_id of the associated line or bar plot.
 template <typename T> IMPLOT_API void PlotErrorBars(const char* label_id, const T* xs, const T* ys, const T* err, int count, int offset=0, int stride=sizeof(T));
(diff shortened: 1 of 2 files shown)


epezent:
Thanks Joel. This is an issue I'm aware of and concerned about.

I recently noticed the negligible difference between non-G and G versions. I was surprised because when I ran tests sometime back, there was quite a bit of difference. Indeed, it does seem that we've slowed down the non-G versions with all the striding/offset/wrap-around code, and other changes.

I haven't focused too much on optimizations yet, as the API is steadily evolving and changing. However, I see us approaching `v1.0` soon, and this is something I want to work on before that point.

I'm torn on how to handle this, because:

1) stride is absolutely necessary for plotting structured data
2) offset and modulo are needed for plotting circular/ring buffers (e.g. for real time plots)

I feel strongly that these need to be supported, if not standard features for all plotting functions. There might be some optimizations we can make to these mechanisms; I have not looked yet. 

Now, whether or not we need `ImPosMod` vs regular modulo is debatable. `ImPosMod` essentially allows you to provide a negative offset. It seemed like an interesting idea at the time, because you could in theory plot RT data scrolling in reverse. In hindsight, that is almost certainly not worth a 30% performance hit. 

We need to look at this more closely, decide what features should be standard, and do more testing. This is definitely something I want to nail down before `v1.0` (at which point the core API will be locked down until an eventual `v2.0`).

On a similar note, I also need to bite the bullet and look at the assembly at some point. My assumption throughout this all has been that the use of the current templated architecture would give us a compile time advantage. It is faster than using the built-in `ImDrawList` API, but I don't know where exactly the gains are coming from. I'm hoping to get to the bottom of it soon, so that I can give feedback to @ocornut. 

Looping in the usual contributors to get their thoughts: @ozlb,  @hoffstadt, @sergeyn, @bear24rw, @PeterJohnson, @jpieper,

sergeyn:
Sorry, I haven't been in the loop of development for a while.  I can see that all getters now have ImPosMod and I believe compilers won't get around it since they don't know offset is not 0.  At the minimum I'd add getters which do not call into ImPosMod when you call with zero 'offset' parameter. You don't necessarily want to  add overloads as you can always check for offset being 0.  In general ImPosMod shouldn't exist since you can always split the data into two regions - before and after wraparound point but that would of course make existing logic slightly more complicated.  As usual, how fast you want to go depends on how much you are willing to give up on maintainability as the code becomes less and less pretty the more perf. features you are applying.

epezent:
Thanks for the feedback @sergeyn! 

>At the minimum I'd add getters which do not call into ImPosMod when you call with zero 'offset' parameter. You don't necessarily want to add overloads as you can always check for offset being 0.

This sounds logical to me. I think we could also do something similar for when `stride = sizeof(T)`, bypassing both the offset and striding logic. We could templatize the `Getters` on a new `Indexer` object which will be chosen at the top of `PlotX` based on the passed values of `offset` and `stride` (e.g. four options: `Indexer`, `IndexOffset`, `IndexerStride`, and `IndexerOffsetStride`). This of course will cause the compile size to explode, so we should think about other options as well. 

> In general ImPosMod shouldn't exist since you can always split the data into two regions - before and after wraparound point but that would of course make existing logic slightly more complicated.

Someone suggested this a long time ago (I think @ocornut), and it's definitely better than doing a modulo on every index. I don't think it would be too tricky to make this work. I like it.

Out of curiosity, here's a really simplistic estimate of the performance we might gain. This is the current implemention of `OffsetAndStride` (ImPosMod + stride) vs simply returning `data[idx]`. Looking at the 100 and 200 marks (i.e. 100k/200k points), it's roughly a 25% gain for me. Considering the majority of my plots (and I assume others as well) are done without offset or stride, it seems that it could be worth the effort.

opt

There are other areas we can focus on as well. And many of other plotting functions (i.e. not PlotLine) are still using the ImDrawList API functions. These need to be migrated to use our faster method.

epezent:
One more thing I want to figure out is where we stand relative to the competition. To be perfectly honest, I don't know if we are fast or slow compared to other plotting APIs. My goal has never been to be the fastest plotting API by any means; I just wanted good plots for ImGui. But now I am genuinely curious how well we stack up to the rest. Calling on @leeoniya for your insight and suggestions for libraries to compare against (I know uPlot is not intended for realtime applications, but you probably know the space better than anyone else).

epezent:
One last thing I want to make abundantly clear: I am a mechanical engineer who is a self taught programmer. I'm only working without about 3 years of real C++ experience. No doubt coding is my guilty pleasure, but I'm probably not the best person to be making decisions regarding rendering optimizations. There are still a lot of things I simply don't understand, so bear with me if I suggest something profoundly stupid.

If we are going to do this thing right, it will require input from more experienced folks than me (if you're out there!)

:)

sergeyn:
Target performance estimation is rather easy as this code is supposed to be memory bandwidth bound (e.g.how much data you write per vertex times the time it takes). If you are less than max memory bandwidth you can achieve on your machine, you have opportunities for improvements.
As for hardware optimizations, you are restricted by ImGui here which dictates that you output textured quads. When at some point in the future custom primitives will be allowed, then un-indexed line-strips (or tri-strips for thick plots) is the way to go.
Maybe it would be handy if you'd add a few benchmarks in your demo code with timings displayed.

JoelLinn:
> I recently noticed the negligible difference between non-G and G versions. I was surprised because when I ran tests sometime back, there was quite a bit of difference.

So was I. There are other non-inlined per-point function calls but not in the order that an additional one wouldn't be perceptible. [This sample](https://godbolt.org/z/nYoh65) shows that for simple getter functions the compiler will not set up a stack frame (depends on compiler arch and calling convention of course) - the overhead then only is like 4 (register)mov instructions and 1 call/ret pair. I'd say that's negligible when the getter is not dropped from L1 cache. This getter is very simple though.
Additionally as @sergeyn said the getter might be I/O bound.

> 1. stride is absolutely necessary for plotting structured data
> 2. offset and modulo are needed for plotting circular/ring buffers (e.g. for real time plots)
>
> I feel strongly that these need to be supported

Agreed.

> ImPosMod essentially allows you to provide a negative offset. It seemed like an interesting idea at the time, because you could in theory plot RT data scrolling in reverse. In hindsight, that is almost certainly not worth a 30% performance hit.

` ImPosMod()` is only part of the equation, `OffsetAndStride()` has a similar contribution. Speaking of which, that function can be optimized since the offset was already clamped between 0 and count in the getter constructor, can't it?

```c++
inline T OffsetAndStride(const T* data, int idx, int count, int offset, int stride) {
    // Now:
    // idx = ImPosMod(offset + idx, count);
    // Faster:
    // idx = (offset + idx) % count;
    // Maybe this (untested) as divisions are really expensive:
    idx += offset;
    if (idx >= count)
        idx -= count;
    
    return *(const T*)(const void*)((const unsigned char*)data + (size_t)idx * stride);
}
```

> At the minimum I'd add getters which do not call into ImPosMod when you call with zero 'offset' parameter. You don't necessarily want to add overloads as you can always check for offset being 0.

If the performance impact of calling a getter for each data point is negligible even for larger getter functions it could be possible to craft templatized getter functions that replace the current getter structs and template instantiations of the plot function itself. That way the compiled code size doesn't explode and there can be template specializations for optimizable cases which can be chosen at runtime. A prettier Version of what [I did with the python wrapper](https://github.com/JoelLinn/py-mahi-gui/blob/master/src/implot.cpp#L53-L140) (`get_getter_func()` will select the correct getter function) to support arbitrary type combinations when only float was supported by implot.

> One more thing I want to figure out is where we stand relative to the competition.

That is an interesting question indeed. ImPlot is special in that it is a realtime and immediate mode plotter. The later means the plot is essentially regenerated from scratch every frame (of course y'all know that). If compared against something like pgfplots (for PFDs, a little bit unfair), we are orders of magnitude faster.
The way lines are drawn in ImGui (2 or 6 triangles depending on the width) means there is a lot of data to be transmitted to the GPU. Interestingly I seem to be either CPU or GPU bound (one of those is at 100%) so the bandwidth seems to be enough.

To plot huge datasets stride is a cheap but dangerous tool (peaks can get lost). It might however be possible to skip a data-point if the distance from the last in pixels is < 1. While this calculation is quick for linear plots, it's penalty might be to high for logarithmic plots. soulthreads/imgui-plot had that implemented but I didn't test it with bigger datasets (like half a million points).

leeoniya:
@epezent benchmarks are super hard. it turns into apples vs oranges almost without exception. there are plenty of often-needed features which can slow down raw perf. eg. scale auto-ranging from visible data (which becomes a 2-pass process), null/gap detection & clipping (oscilloscope-type applications don't need this), configurable line width (bare webgl can only do 1px widths). many charting libs don't support timezones & dst which could affect perf with temporal scales. interactivity perf is another huge caveat (series toggle, live legend display, streaming/redraw).

some of the wrbgl libs i mention in uPlot's intro [1] explicitly opt out of supporting variable line width, saying it would greatly complicate the shaders and affect perf. others say that auto scale ranging is too much ovehead to add. still others say that data gaps should not be in the user's data and should be annotated using another method.

i'm not sure how much deep insight i can provide beyond web-sphere charting libs. i'm betting the perf of native code would be faster than the same transpiled to WASM/webgl.

i guess https://huww98.github.io/TimeChart/docs/performance is another useful bench with linewidth support.

i've recently stumbled upon https://github.com/38/plotters.

typically webgl's raw perf is good (though it always turns my laptop into a stovetop & jet engine). the devil is in the details.

sorry, i dont think any of the above is super useful/actionable.

[1] https://github.com/leeoniya/uPlot#introduction

leeoniya:
> To plot huge datasets stride is a cheap but dangerous tool (peaks can get lost). It might however be possible to skip a data-point if the distance from the last in pixels is < 1. While this calculation is quick for linear plots, it's penalty might be to high for logarithmic plots. soulthreads/imgui-plot had that implemented but I didn't test it with bigger datasets (like half a million points).

this is correct. uPlot does a draw-time decimation that accumulates entry/high/low/exit y values for every pixel. this is the only way to do it properly, as any other kinds of sampling will almost certainl drop critical data. https://knowledge.ni.com/KnowledgeArticleDetails?id=kA00Z0000019YLKSA2&l=en-US

JoelLinn:
> (bare webgl can only do 1px widths)

guess what, ImGui only knows triangles and uses two of them for a line segment that's 1px wide and 6 of them for everything >1px if AA is enabled (ImPlot doesn't do that though).  So in our case its 4 vertices per data point (1px width), WebGL should be fine.

> uPlot does a draw-time decimation that accumulates entry/high/low/exit y values for every pixel.

I have to look into those decimation algorithms, checking for `(x2 - x1)^2 + (y2 - y1)^2 < 1^2` didn't sound too bad to be honest (but yeah in some cases, it can be inaccurate)

leeoniya:
another thing to try is LTTB downsampling:

https://github.com/sveinn-steinarsson/flot-downsample

ozlb:
Downsampling is something that i'm considering since beginning and probably there are benefits in "dense" plots; it can be implemented "externally" in demo to evaluate when it starts to give benefits (C++ LTTB https://github.com/parkertomatoes/lttb-cpp looks nice)
For realtime plots I'm using custom getter function in order to iterate a portion of big circular data buffer and that portion amount is pre-calculated as x axis span divided by sample time (and a factor to compensate non constant sampling); when real time plot is "paused" (but acquiring is still running) i search binary the x axis samples locations (only on x axis changes) and plot that buffer portion. The result is that even if buffer holds huge amount of points, only the "culled" samples are looped by plot.

ozlb:
Actually I successfully implemented LTTB (or to say better the cpp with some adjustments) and forse long term and dense plot there are huge benefits without significant quality compromises.
It will follow a ImPlot::Demo as soon as possible.

ozlb:
Screenshot comparative.
There are some differences (i.e. at peek of orange signal), but in mine application more than acceptable.

Screenshot from 2020-10-01 00-28-34
Screenshot from 2020-10-01 00-28-48

ozlb:
https://github.com/ozlb/implot/commit/a5543a1a8ab8caa84e5843e54647effaa905ef43

lastthyme (approved):

epezent:
I've been experimenting with optimizing line rendering lately. First off, our custom implementation of `PlotLine` is already 50% faster than calling directly into ImGui's drawing API (note this is with ImGui AA turned OFF to make things fair). That's quite significant!

Here, the solid marker lines are total frame time, while the open marker line is call time into the plotting function only, e.g. `ImPlot::PlotLine`

image

By eliminating unnecessary offsetting/striding calculations when indexing (as this PR suggested), I am able to almost half time spent inside of `ImPlot::PlotLine`. Unfortunately, this doesn't necessarily translate to much improvement in overall frame time (only ~15%).

image

Out of curisoity, I looked into using AVX instructions to triangulate 8 line segments at a time. Surprisingly this offers almost not improvement, either in self time or frame time.

epezent:
Closing this as #270 address this issue and improves line rendering performance by a ~45%

PR #123, opened 2020-09-22
CalcTextColor() fix

ozlb:
Returns white or black text given background color fix


epezent:
There's some other code in this PR that's not related. I will update the 0.5 manually.

PR #125, opened 2020-10-05
Add CMake integration

seanballais:
I am using ImPlot in a project of mine. My project uses CMake, and having CMake support in dependencies that I obtain from repositories makes it easier for me to use them in my project. As such, I have added a `CMakeLists.txt` to my fork of ImPlot so that I can integrate it more easily to my project. I am writing this PR in the hopes that my modifications might be helpful to other developers.

Developers can integrate ImPlot to their CMake project by adding the following lines in a `CMakeLists.txt` file:

```cmake
include_directories(path/to/implot/)
add_subdirectory(path/to/implot/)
```

And then link ImPlot to their executable by doing

```cmake
target_link_libraries(<target executable>
    implot
    <other libraries>)
```

if they want to link the static version of ImPlot, or

```cmake
target_link_libraries(<target executable>
    implot_shared
    <other libraries>)
```

if they want to the shared version.

I have added three options in the `CMakeLists.txt` file that can be toggled on or off by setting their values to either `ON` or `OFF`, respectively:

* `BUILD_IMPLOT_SHARED` - builds a shared library version of ImPlot; turned on by default
* `BUILD_IMPLOT_STATIC` - builds a static library version of ImPlot; turned on by default
* `BUILD_IMPLOT_DEMO` - includes the demo in the library; turned off by default

I've also updated the README to include the documentation for CMake integration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,30 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(implot)
+
+option(BUILD_IMPLOT_SHARED "Build implot shared library" ON)
+option(BUILD_IMPLOT_STATIC "Build implot static library" ON)
+option(BUILD_IMPLOT_DEMO "Build implot demo" OFF)
+
+SET(IMPLOT_SOURCES
+    implot.cpp
+    implot_items.cpp)
+
+if (BUILD_IMPLOT_DEMO)
+    message(STATUS "Building with implot demo.")
+    SET(IMPLOT_SOURCES
+        ${IMPLOT_SOURCES}
+        implot_demo.cpp)
+endif ()
+
+if (BUILD_IMPLOT_SHARED)
+    message(STATUS "Building implot as a shared library.")
+    add_library(implot_shared SHARED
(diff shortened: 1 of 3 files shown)


epezent:
Hi! Thank you for submitting this PR. I'm sure some may find it useful. However, I will not be merging this yet as:

1) ImGui does not yet support CMake officially, so it feels strange for us to
2) Adding ImPlot sources manually to a target in CMake is about as much code as what's above. There's not much benefit that I can see. 
3) I suspect that people using ImPlot either add its sources directly to their own, or add them to their ImGui library complication. Compiling ImPlot as a separate lib seems like overkill. 

If and when ImGui provides support for CMake, I'll reopen the issue. To be clear, I use CMake religously and I don't see much benefit to supporting it until ImGui does. Follow their on-going discussion here: https://github.com/ocornut/imgui/pull/1713 

Also, just as an FYI to you and others: you should probably never link ImGui or ImPlot dynamically as both are quite call heavy.

seanballais:
Hey, @epezent! Thanks for the feedback! I can see why my PR would be overkill at the moment in general.

PR #140, opened 2020-11-04
Option for not rendering the title of the graph.

npostma:
Option for not rendering the title of the graph.

I needed a small form factor bar chart. Setting the flag results in:

image


epezent:
This is already possible by prepending "##" to the title/ID string. Is there a specific reason you need a flag instead?

npostma:
Ah no. I did not know this was an option. And i couldn't find one to hide. So I've added a less magical implementation i guess. This can be closed if it has no benefit over ##

epezent:
I don't suppose it hurts. It might make turning the title on/off programmatically easier in some use cases.

PR #142, opened 2020-11-08
Don't redeclare IM_NORMALIZE2F_OVER_ZERO unnecessarily

FredTheDino:
I got annoyed at the warnings I got compiling with GCC 10.2.0. 

If there's a better way to solve this let me know, but I figured
this was the simplest and best way. :D

### DIFF
--- a/implot_items.cpp
+++ b/implot_items.cpp
@@ -32,7 +32,7 @@
 #define SQRT_1_2 0.70710678118f
 #define SQRT_3_2 0.86602540378f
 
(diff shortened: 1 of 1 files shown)


epezent:
Can you just rename the macro in ImPlot to `IMPLOT_NORMALIZE2F_OVER_ZERO` throughout? I don't want to risk the chance that the define is not what we expect.

FredTheDino:
Very sound reasoning! Updated! :D

Issue #53, opened 2020-06-07
Visualisation Improvement ideas/fixes


sergeyn:
1.) Minor ticks should not suddenly become major ticks, but gradually blend into major ticks. New minor ticks should blend in out of completely transparent ones. This will look much nicer when you do zooming, especially very rapid zooming. The point is not to have sudden changes in an image
2.) Add very mild animation on zooming - instead of sudden changes in zoom, zoom in gradually, starting at very fast speed, and slowing down when reaching zoom target, play the whole animation within 0.5-0.8second so it is still as sharp and responsive and not annoying. Also put it on a flag 
3.) When zooming in, amount of digits for each tick grows and at some point they start to overlap. 
![image](https://user-images.githubusercontent.com/668090/83968779-d1359a80-a8cb-11ea-8544-22839295b8b0.png)
I suggest skipping outputting text for ticks which will overlap with previous tick, with preference for major ticks of course.

ozlb:
> https://github.com/epezent/implot/issues/1 What if instead we made the edges and corners of Query dragable/resizable after it is created? If the query was initially created using the Shift key (to expand vertical), then it would be the exact same behavior as far as I can tell. It also has the added benefit of working in the Y direction when needed. I'd just rather not have two similar tools to manage and upkeep.

I will re-trigger Query improvements, since I'm still updating my own local version with my "beloved" QueryX and main reason is that for my needs Query is not covering:
**- resizing:** Query cannot be resized or I'm not able to do it (as per @epezent post here above)
**- lock on x or y expansion:** when expanded, Query dragable in all directions I think it's something non needed: it should drag only horizontally if expanded vertically and vice-versa. It will be nice option to re-scale based on x axis range changes if expanded vertically.
**- show/hide API:** actual 'ImPlotFlags_Query' can condition only the user interaction but I would like to control the visibility. Actually about this, there is a "bug" I think, because once Query has been plot, if you deselect the "Settings -> Query" it will be visible forever.
**- force vertically expansion programmatically:** I would like to "force" this behavior
**- graphically presence:** fill the rectangle it's in my opinion non necessary if the selection is expanded and have only the lines with same ticker line where is crossing axis, so as per QueryX
![QueryX](https://user-images.githubusercontent.com/3250876/83970304-5ec9b800-a8d5-11ea-9113-ab77b34d127b.gif)

epezent:
[...]

PR #144, opened 2020-11-09
AlignedPlots: Y/X axis padding over multiple plots

ozlb:
To have same Y axis padding, `ImPlot::BeginSubPlots()` must be called before 1st `ImPlot::BeginPlot()` and `ImPlot::EndSubPlots()` must be called after last `ImPlot::EndPlot()`. 
Peek 2020-11-10 22-44

https://github.com/epezent/implot/issues/53#issuecomment-723526658


ozlb:
@epezent 
This is a working draft; final version requires a SubPlot ID pool to have the possibility to handle multiple sub plots. Thoughts?

epezent:
@ozlb, thanks for looking into this. A few questions:

- Yes, I think you'll need a pool which stores padding information. You'll need to add either a string or int ID to the function signature to do this. I suggest adding a new struct to implot_internal.h that stores all the necessary state information. 
- In your example, the top plot has the widest labels, and so it propagates to the bottom plot in one frame. In the reverse case, the top plot won't update until the next frame. How noticeable is the delay? Can you show an example where the bottom plot has the widest labels?
- Can you show an example with auxiliary y-axis? 
- What happens when a plot has an axis title?
- What happens when a plot has neither an axis title or tick labels?
- After thinking about it, I think we might want to change the name of this function to `BeginAlignedLabels`/`EndAlignedLabels`. I think `BeginSubplot` would need to offer additional functionality to warrant the name, e.g. similar to [matplotlib subplots](https://matplotlib.org/3.3.2/api/_as_gen/matplotlib.pyplot.subplot.html). That of course is also an option, but I think plots inside of ImGui Tables pretty much covers most use cases already. (e.g. grids of plots).

ozlb:
- new signature with ID: i already implemented and i will post updates asap
- bottom to top propagation: the delay is only for a single cycle because on the next one if padding will not change it will match the saved. I can’t see delays. I will capture and share asap
- I’m regularly testing with 3 y axis, no problem. I will capture and share asap
- axis labels or tick labels and ticks are calculated before at plot bb (3), so are included. I will capture and share asap
- i renamed in BeginAlignedPlots

Thanks for your help

epezent:
Awesome! One other thing: can we add support for vertical alignment when `ImGui::SameLine()` is used?

ozlb:
Peek 2020-11-11 23-32

ozlb:
> Awesome! One other thing: can we add support for vertical alignment when `ImGui::SameLine()` is used?

I changed signature of function to handle vertically/horizontally using already existing enum `ImPlotOrientation`. 

Peek 2020-11-12 00-26

Thoughts?

epezent:
> I changed signature of function to handle vertically/horizontally using already existing enum `ImPlotOrientation`. 

Love it! Another thought: 

If we changed `ImPlotOrientation` to be a flag, then it could be possible to pass `ImPlotOrientation_Horizontal | ImPlotOrientatoin_Vertical` for 2D grids of plots. Is that overkill? 

Otherwise this looks great, I'll try to take a look tonight or tomorrow.

ozlb:
> If we changed `ImPlotOrientation` to be a flag, then it could be possible to pass `ImPlotOrientation_Horizontal | ImPlotOrientatoin_Vertical` for 2D grids of plots. Is that overkill?

signature is ok, but to use as flag, enum `ImPlotOrientation_` need to be changed and start from 1.

```
// Used to orient items on a plot (e.g. legends, labels, etc.) or align plots
enum ImPlotOrientation_ {
    ImPlotOrientation_None       = 0,      // 
    ImPlotOrientation_Horizontal = 1 << 0, // left/right
    ImPlotOrientation_Vertical   = 1 << 1, // up/down
};

```
I cross check where is used for legends, labels and should not be problematic.

ozlb:
Ready for your final approval / refinements / decorations.

epezent:
@ozlb , sorry I haven't addressed this. I wanted to first merge organizational changes I've been making in `implot_internal.h` with the addition of `ImPlotFlags_Equal`. That's done, and so this needs to have a few conflicts resolved.

ozlb:
I will resolve the conflicts and update PR.

ozlb:
Done

Mrgoodbytes13:
Would it be possible to also add crosshair alignment for plot groups?
ie:
aligned

ozlb:
interesing, but before I need to sync to latest version so hopefully it will be merged into master.

yamen:
I came here to look for a way to do this, and here's a PR with exactly what I need! This is fantastic work.

I think crosshair alignment should be covered separately, as it only depends on X-value linking, which already has a working solution. Once we get visual alignment of X-axis, the rest will work fine.

epezent:
@yamen and @ozlb , I hope to integrate this soon. I know this a feature that many will enjoy. Sorry for the delay.

arximboldi:
Oh yes!

dgm3333:
Can I vote for the ozlb code getting pulled into the main branch - I've got a couple of plots I ended up aligning by modifying the axis text :-( but this would be much better.

dgm3333:
Incidentally (as it's related to alignment)
There also seems to be an oddity I can't consistently replicate where if you don't call the SetNextPlotLimitsX before EVERY linked plot the xmin gets set to 0 for all of them (but it seems sometimes OK not to??).
Also I originally assumed you didn't have to manually call SetNextPlotLimitsX (as it's not in the original demo script), but this isn't the case - should the link ALSO call SetNextPlotLimitsX since xmin/max/etc are being set to a value?

if (ImGui::CollapsingHeader("Linked Axes")) {
    static double xmin = 10, xmax = 11, ymin = 0, ymax = 1;
    static bool linkx = true, linky = true;
    int data[2] = {0,1};
    int series[2] = { 10,11 };
    ImGui::Checkbox("Link X", &linkx);
    ImGui::SameLine();
    ImGui::Checkbox("Link Y", &linky);
    ImPlot::LinkNextPlotLimits(linkx ? &xmin : NULL , linkx ? &xmax : NULL, linky ? &ymin : NULL, linky ? &ymax : NULL);
    //ImPlot::SetNextPlotLimitsX(xmin, xmax, ImGuiCond_Once);
    if (ImPlot::BeginPlot("Plot A")) {
        ImPlot::PlotLine("Line",series,data,2);
        ImPlot::EndPlot();
    }
    ImPlot::LinkNextPlotLimits(linkx ? &xmin : NULL , linkx ? &xmax : NULL, linky ? &ymin : NULL, linky ? &ymax : NULL);
    //ImPlot::SetNextPlotLimitsX(xmin, xmax, ImGuiCond_Once);
    if (ImPlot::BeginPlot("Plot B")) {
        ImPlot::PlotLine("Line",series, data,2);
        ImPlot::EndPlot();
    }
}

(using latest imgui and implot compiling win64 c++20 under Visual Studio)

epezent:
Merged as a part of #203

PR #145, opened 2020-11-15
Fixed bounds computation for heatmap labels

4bb4:
Hi there, 

While creating bindings for the heatmap plots, I noticed that the labels don't seem to get drawn in the right place when I change the bounds. The (-1,-1) and (0,0) case shows up properly, like in your demo code:

image

however, if I for example change the bounds to (-1,-1) and (1,1), I get this picture:

image

If I make the change of this PR, it shows up as expected:

image

Note that I didn't think about this for a long time, I just copied the computation you use for the `px` of  points themselves and tried out a few cases. I'm also trying this out from the Rust bindings (https://github.com/4bb4/implot-rs says hi!), so there is a chance that I messed up on the integration side and didn't find the issue. If that is the case, apologies in advance :+1:

### DIFF
--- a/implot_items.cpp
+++ b/implot_items.cpp
@@ -1684,7 +1684,7 @@ void RenderHeatmap(Transformer transformer, ImDrawList& DrawList, const T* value
             for (int c = 0; c < cols; ++c) {
                 ImPlotPoint p;
                 p.x = bounds_min.x + 0.5*w + c*w;
-                p.y = bounds_min.y + 1 - (0.5*h + r*h);
+                p.y = bounds_max.y - (0.5*h + r*h);
                 ImVec2 px = transformer(p);
                 char buff[32];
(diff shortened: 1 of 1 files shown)


epezent:
I noticed the issue as well while working on the new `PlotHistograms2D`. I'll cross check your solution with mine and get this fixed relatively soon.

4bb4:
Sorry about the mess with the PR closing, I committed with the wrong account and tried to remove and re-do it. Doesn't seem possible to delete that commit entirely though.

epezent:
Don't worry, it's just one line of code so I can add it manually.

4bb4:
> get this fixed relatively soon. 

Great, thanks - I'm not actually using heatmaps at the moment, I'm just adding more and more coverage of the API and heatmaps were the item of the week :) I'll add an issue so I can subscribe to it, should have done that anyway.

PR #149, opened 2020-11-29
render x-axis drag/drop hover

phkehl:
Similar to your Drag and Drop demo I have a drag and drop plot setup where the user can select from a number of variables to plot. I want them to be able to drop one variable on the x-axis (and use that as the x variable). For that I thought highlighting the x-axis like the y-axes would be good.
I'm not insisting on this pull request being merged. I can maintain that patch in my project, no problem. But I thought it could be useful thing to have.
I'm not sure how that fits your concept. Perhaps you'd prefer this to be an option (ImPlotFlags or ImPlotAxisFlags perhaps)?
Please advise.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -2221,6 +2221,16 @@ void EndPlot() {
         }
     }
 
+    // render x-axis drag/drop hover
+    if (plot.XAxis.Present && ImGui::IsDragDropPayloadBeingAccepted()) {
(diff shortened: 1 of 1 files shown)


epezent:
I can't think of any issues this would cause, so I'm fine merging your PR. I'll review and test it when I have a chance.

PR #154, opened 2020-12-04
Center constant data when fitting.

mindv0rtex:
This is an extremely simple fix, I am open to changing it. The reason the current logic does not work is that if the `plot.XAxis.Range.Max` happens to be zero, nothing will happen. I reasoned that we don't really need to shift ranges by some percentage of the max value, we just center it.


epezent:
I see, that makes sense. Since we have to pick an arbitrary amount to add to the min/max, I agree that +/-0.5 for a single unit span is probably the most logical choice. Merging.

PR #157, opened 2020-12-08
Fix ID collisions in the demo

rokups:
Several `CollapsingHeader()` use same label as `BeginPlot()`. This creates an ID collision, which confuses Dear ImGui testing engine. Added `##` in `CollapsingHeader()` label are included in ID calculation, but discarded when rendering. This gives affected `CollapsingHeader()` a different ID and solves ID collision.


epezent:
Thanks @rokups!

PR #163, opened 2021-01-06
Allow second and third axes to have labels

mindv0rtex:
This is how the new demo looks like:
Screen Shot 2021-01-05 at 22 25 01


epezent:
Yan, this is excellent! I made a few minor edits and added `ImPlotAxisFlags_NoLabel` so that axis labels can be hidden even when a string is passed to `BeginPlot` (useful for hiding labels from the context menu).

Merging!

PR #239, opened 2021-06-03
Fix for ImPool change in Dear ImGui version 18303+

ocornut:
Fix for 
https://github.com/ocornut/imgui/commit/c6c82b9f1de6ef3e298050656a1cf0a499d4e808

As previously, because we don't want users to have to perfectly sync ImPlot and ImGui, I did two things:

- Code in ImPlot can support older versions.

- Dear ImGui itself has a graceful redirection functions
```
#ifndef IMGUI_DISABLE_OBSOLETE_FUNCTIONS
    int   GetSize()  { return GetBufSize(); } // For ImPlot: should use GetBufSize() from (IMGUI_VERSION_NUM >= 18303)
#endif
```
I expect to remove this in a few versions

- FYI: that iterating ImPool with GetSize() + GetByIndex() would be incorrect if you ever used ImPool::Remove() (which you aren't). Commit above aims to clarify that.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -78,9 +78,14 @@ You can read releases logs https://github.com/epezent/implot/releases for more d
 #define sprintf sprintf_s
 #endif
 
-// Support for pre-1.82 version. Users on 1.82+ can use 0 (default) flags to mean "all corners" but in order to support older versions we are more explicit.
+// Support for pre-1.82 versions. Users on 1.82+ can use 0 (default) flags to mean "all corners" but in order to support older versions we are more explicit.
 #if (IMGUI_VERSION_NUM < 18102) && !defined(ImDrawFlags_RoundCornersAll)
 #define ImDrawFlags_RoundCornersAll ImDrawCornerFlags_All
+#endif
+
+// Support for pre-1.84 versions. ImPool's GetSize() -> GetBufSize()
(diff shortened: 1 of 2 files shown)


epezent:
Thanks Omar!

ocornut:
Note I made another fix:
https://github.com/ocornut/imgui/commit/94b680e83052b9ff2e877360309020b72db057f2

GetBufSize() renamed to GetMapSize(), and added an actual GetBufSize().
Should NOT affect ImPool at all: in your situation using GetBufSize() or GetMapSize() will give you same value and the earlier is more correct considering to use GetByIndex(n) afterwards.
Only if you start removing elements you may need to rework your iteration on ImPool<> data later.

Max1412:
Hi, I'm trying to use ImPlot with Dear ImGui Version with IMGUI_VERSION_NUM 18203 and the define fix for GetBufSize seems to work in all instances except in implot_internal.h line 785, since it is not defined in this file. Am I doing something wrong with regards to the include order? Or is it just missing in this file?

Max1412:
I was on a version of master two commits earlier than when it got added to implot_internal.h and I just upgraded and it works perfectly fine now. Thanks for the fast reply!

epezent:
No problem! This was an oversight on my part when merging Subplots.

PR #254, opened 2021-06-20
Add GPU-accelerated heatmaps

marcizhu:
As promised in #219, I managed to implement GPU-accelerated heatmaps! I haven't done any serious benchmarks but the performance gains I've seen in some applications are quite significant.

This PR is pretty much ready to be merged, but there are some small issues that should be addressed:
1. OpenGL doesn't support `double` nor `*int64_t` (aka `ImU64`/`ImS64`). If this kind of data is passed to a heatmap, it is internally converted to `float` or `ImU32`/`ImS32` (respectively).
2. When using OpenGL acceleration, the interpolation tables for heatmaps are unnecessary. Maybe we could only enable them in case this optimization is disabled
3. Right now I use the same trick ImGui's `main.cpp` file uses to access OpenGL functions, which is rely on a macro definition to include one of the many supported libraries. It is possible to implement our own OpenGL loader so that we don't rely on any external library (ImGui will still depend on that library though), but that would add a bit of complexity maybe.
4. Log-Log, Lin-Log and Log-Lin axis are not yet implemented.

I tried to implement this as straightforward as possible, and with as little modifications to the 'core' of the library as possible. Also, all the implementation details are inside the file `backends/implot_opengl.cpp`. If the macro `IMPLOT_ENABLE_OPENGL3_ACCELERATION` is defined, then that file will compile and the library will use the shader-accelerated code. If not, it will default to the current method.

I will update this PR once I have some benchmarks to show you, but the performance gains should be significant :)

Any comments, optimizations, questions or changes are always welcome :D

### DIFF
--- a/backends/implot_backend.h
+++ b/backends/implot_backend.h
@@ -26,6 +26,12 @@
 
 #include "../implot.h"
 
+#ifdef IMPLOT_BACKEND_ENABLE_OPENGL3
+	#define IMPLOT_BACKEND_ENABLED
+	#define IMPLOT_BACKEND_HAS_HEATMAP
+	#define IMPLOT_BACKEND_HAS_COLORMAP
+#endif
+
 namespace ImPlot {
 namespace Backend {
 
@@ -136,17 +142,25 @@ IMPLOT_API void BustPlotCache();
 /** @brief Bust item cache. Called from @ref ImPlot::BustItemCache() */
 IMPLOT_API void BustItemCache();
 
+}
+}
 
-#if !defined(IMPLOT_ENABLE_OPENGL3_ACCELERATION)
+namespace ImPlot {
+namespace Backend {
 
-// Dummy implementation for backend functions
+#ifndef IMPLOT_BACKEND_ENABLED
+
 inline void* CreateContext() { return nullptr; }
 inline void DestroyContext() {}
 
-inline void AddColormap(const ImU32*, int, bool) {}
-
 inline void BustPlotCache() {}
 inline void BustItemCache() {}
+
+#endif
+
+#ifndef IMPLOT_BACKEND_HAS_COLORMAP
+
+inline void AddColormap(const ImU32*, int, bool) {}
 
 #endif
 
--- a/backends/implot_backend.h
+++ b/backends/implot_backend.h
@@ -35,6 +35,10 @@
 namespace ImPlot {
 namespace Backend {
 
+//-----------------------------------------------------------------------------
+// [SECTION] Misc backend functions
+//-----------------------------------------------------------------------------
+
 /**
  * @brief Struct to hold backend-related context data
  *
@@ -62,6 +66,16 @@ IMPLOT_API void* CreateContext();
  */
 IMPLOT_API void DestroyContext();
 
+/** @brief Bust plot cache. Called from @ref ImPlot::BustPlotCache() */
+IMPLOT_API void BustPlotCache();
+
+/** @brief Bust item cache. Called from @ref ImPlot::BustItemCache() */
+IMPLOT_API void BustItemCache();
+
+//-----------------------------------------------------------------------------
+// [SECTION] Colormap functions
+//-----------------------------------------------------------------------------
+
 /**
  * @brief Add a colormap
  *
@@ -73,6 +87,10 @@ IMPLOT_API void DestroyContext();
  *               not (`true`)
  */
 IMPLOT_API void AddColormap(const ImU32* keys, int count, bool qual);
+
+//-----------------------------------------------------------------------------
+// [SECTION] Heatmap functions
+//-----------------------------------------------------------------------------
 
 /**
  * @brief Set heatmap data
@@ -135,12 +153,6 @@ IMPLOT_API void SetHeatmapData(int plotID, const ImU64*  values, int rows, int c
 IMPLOT_API void RenderHeatmap(
 	int plotID, ImDrawList& DrawList, const ImVec2& bounds_min, const ImVec2& bounds_max,
 	float scale_min, float scale_max, ImPlotColormap colormap);
-
-/** @brief Bust plot cache. Called from @ref ImPlot::BustPlotCache() */
-IMPLOT_API void BustPlotCache();
-
-/** @brief Bust item cache. Called from @ref ImPlot::BustItemCache() */
-IMPLOT_API void BustItemCache();
 
 }
 }
--- a/backends/implot_backend.h
+++ b/backends/implot_backend.h
@@ -24,156 +24,25 @@
 
 #pragma once
 
-#include "../implot.h"
-
-#ifdef IMPLOT_BACKEND_ENABLE_OPENGL3
-	#define IMPLOT_BACKEND_ENABLED
-	#define IMPLOT_BACKEND_HAS_HEATMAP
-	#define IMPLOT_BACKEND_HAS_COLORMAP
+#if defined(IMPLOT_BACKEND_ENABLE_OPENGL3)
+	#include "implot_impl_opengl3.h"
+#elif defined(IMPLOT_BACKEND_ENABLE_METAL)
+	#include "implot_impl_metal.h"
 #endif
-
-namespace ImPlot {
-namespace Backend {
-
-//-----------------------------------------------------------------------------
-// [SECTION] Misc backend functions
-//-----------------------------------------------------------------------------
-
-/**
- * @brief Struct to hold backend-related context data
- *
- * A backend may store in this struct any data it needs, with no constraints. A
- * pointer to this struct will be stored inside ImPlot's context and can be
- * accessed at any time. This pointer will be set to the returned value of @ref
- * CreateContext(). All resources held by this struct must be freed inside @ref
- * DestroyContext().
- */
-struct ContextData;
-
-/**
- * @brief Create backend context
- *
- * Creates and intializes the backend context. The returned pointer will be saved
- * in ImPlot's context and can be accessed later.
- */
-IMPLOT_API void* CreateContext();
-
-/**
- * @brief Destroy backend context
- *
- * Destroys and frees any memory or resources needed by the backend. After this
- * call returns, no more calls to any backend function will be performed.
- */
-IMPLOT_API void DestroyContext();
-
-/** @brief Bust plot cache. Called from @ref ImPlot::BustPlotCache() */
-IMPLOT_API void BustPlotCache();
-
-/** @brief Bust item cache. Called from @ref ImPlot::BustItemCache() */
-IMPLOT_API void BustItemCache();
-
-//-----------------------------------------------------------------------------
-// [SECTION] Colormap functions
-//-----------------------------------------------------------------------------
-
-/**
- * @brief Add a colormap
- *
- * Adds a colormap to be handled by the backend.
- *
- * @param keys   Colors for this colormap, in RGBA format
- * @param count  Number of colors in this colormap
- * @param qual   Qualitative: whether the colormap is continuous (`false`) or
- *               not (`true`)
- */
-IMPLOT_API void AddColormap(const ImU32* keys, int count, bool qual);
-
-//-----------------------------------------------------------------------------
-// [SECTION] Heatmap functions
-//-----------------------------------------------------------------------------
-
-/**
- * @brief Set heatmap data
- *
- * Sets the data of the heatmap with the given plot ID.
- *
- * @param plotID  ID of the heatmap to update. This ID is unique, but it is not
- *                continuous nor always positive.
- * @param values  Data of the heatmap to be set.`values[0]` corresponds with the
- *                top-left corner of the data.
- * @param rows    Number of rows of this heatmap
- * @param cols    Number of columns of this heatmap
- */
-IMPLOT_API void SetHeatmapData(int plotID, const ImS8*   values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImU8*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImU8*   values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImS16*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImS16*  values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImU16*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImU16*  values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImS32*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImS32*  values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImU32*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImU32*  values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const float*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const float*  values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const double*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const double* values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImS64*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImS64*  values, int rows, int cols);
-
-/** @copydoc SetHeatmapData(int,const ImU64*,int,int) */
-IMPLOT_API void SetHeatmapData(int plotID, const ImU64*  values, int rows, int cols);
-
-/**
- * @brief Render heatmap
- *
- * Renders the heatmap by submitting the appropriate commands to the current
- * draw list.
- *
- * @param plotID      ID of the heatmap to be rendered
- * @param DrawList    Draw list where to submit the render commands
- * @param bounds_min  Minimum bounds of the heatmap (without clipping)
- * @param bounds_max  Maximum bounds of the heatmap (without clipping)
- * @param scale_min   Minimum value of the heatmap
- * @param scale_max   Maximum value of the heatmap
- * @param colormap    Colormap to be used when rendering this heatmap
- *
- * @note There might be values greater than `scale_max` or lower than `scale_min`.
- *       The shader used for rendering should clamp this values appropriately.
- */
-IMPLOT_API void RenderHeatmap(
-	int plotID, ImDrawList& DrawList, const ImVec2& bounds_min, const ImVec2& bounds_max,
-	float scale_min, float scale_max, ImPlotColormap colormap);
-
-}
-}
 
 namespace ImPlot {
 namespace Backend {
 
 #ifndef IMPLOT_BACKEND_ENABLED
+	inline void* CreateContext() { return nullptr; }
+	inline void DestroyContext() {}
 
-inline void* CreateContext() { return nullptr; }
-inline void DestroyContext() {}
-
-inline void BustPlotCache() {}
-inline void BustItemCache() {}
-
+	inline void BustPlotCache() {}
+	inline void BustItemCache() {}
 #endif
 
 #ifndef IMPLOT_BACKEND_HAS_COLORMAP
-
-inline void AddColormap(const ImU32*, int, bool) {}
-
+	inline void AddColormap(const ImU32*, int, bool) {}
 #endif
 
 }
--- a/backends/implot_gpu.h
+++ b/backends/implot_gpu.h
@@ -1,0 +1,154 @@
+// MIT License
+
+// Copyright (c) 2021 Evan Pezent
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// ImPlot v0.10 WIP
+
+#pragma once
+
+#include "../implot.h"
+
+namespace ImPlot {
+namespace Backend {
+
+/**
+ * @brief Struct to hold backend-related context data
+ *
+ * A backend may store in this struct any data it needs, with no constraints. A
+ * pointer to this struct will be stored inside ImPlot's context and can be
+ * accessed at any time. This pointer will be set to the returned value of @ref
+ * CreateContext(). All resources held by this struct must be freed inside @ref
+ * DestroyContext().
+ */
+struct ContextData;
+
+/**
+ * @brief Create backend context
+ *
+ * Creates and intializes the backend context. The returned pointer will be saved
+ * in ImPlot's context and can be accessed later.
+ */
+IMPLOT_API void* CreateContext();
+
+/**
+ * @brief Destroy backend context
+ *
+ * Destroys and frees any memory or resources needed by the backend. After this
+ * call returns, no more calls to any backend function will be performed.
+ */
+IMPLOT_API void DestroyContext();
+
+/**
+ * @brief Add a colormap
+ *
+ * Adds a colormap to be handled by the backend.
+ *
+ * @param keys   Colors for this colormap, in RGBA format
+ * @param count  Number of colors in this colormap
+ * @param qual   Qualitative: whether the colormap is continuous (`false`) or
+ *               not (`true`)
+ */
+IMPLOT_API void AddColormap(const ImU32* keys, int count, bool qual);
+
+/**
+ * @brief Set heatmap data
+ *
+ * Sets the data of the heatmap with the given plot ID.
+ *
+ * @param plotID  ID of the heatmap to update. This ID is unique, but it is not
+ *                continuous nor always positive.
+ * @param values  Data of the heatmap to be set.`values[0]` corresponds with the
+ *                top-left corner of the data.
+ * @param rows    Number of rows of this heatmap
+ * @param cols    Number of columns of this heatmap
+ */
+IMPLOT_API void SetHeatmapData(int plotID, const ImS8*   values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImU8*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImU8*   values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImS16*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImS16*  values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImU16*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImU16*  values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImS32*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImS32*  values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImU32*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImU32*  values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const float*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const float*  values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const double*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const double* values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImS64*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImS64*  values, int rows, int cols);
+
+/** @copydoc SetHeatmapData(int,const ImU64*,int,int) */
+IMPLOT_API void SetHeatmapData(int plotID, const ImU64*  values, int rows, int cols);
+
+/**
+ * @brief Render heatmap
+ *
+ * Renders the heatmap by submitting the appropriate commands to the current
+ * draw list.
+ *
+ * @param plotID      ID of the heatmap to be rendered
+ * @param DrawList    Draw list where to submit the render commands
+ * @param bounds_min  Minimum bounds of the heatmap (without clipping)
+ * @param bounds_max  Maximum bounds of the heatmap (without clipping)
+ * @param scale_min   Minimum value of the heatmap
+ * @param scale_max   Maximum value of the heatmap
+ * @param colormap    Colormap to be used when rendering this heatmap
+ *
+ * @note There might be values greater than `scale_max` or lower than `scale_min`.
+ *       The shader used for rendering should clamp this values appropriately.
+ */
+IMPLOT_API void RenderHeatmap(
+	int plotID, ImDrawList& DrawList, const ImVec2& bounds_min, const ImVec2& bounds_max,
+	float scale_min, float scale_max, ImPlotColormap colormap);
+
+/** @brief Bust plot cache. Called from @ref ImPlot::BustPlotCache() */
+IMPLOT_API void BustPlotCache();
+
+/** @brief Bust item cache. Called from @ref ImPlot::BustItemCache() */
+IMPLOT_API void BustItemCache();
+
+
+#if !defined(IMPLOT_ENABLE_OPENGL3_ACCELERATION)
+
+// Dummy implementation for backend functions
+inline void* CreateContext() { return nullptr; }
+inline void DestroyContext() {}
+
+inline void AddColormap(const ImU32*, int, bool) {}
+
+inline void BustPlotCache() {}
+inline void BustItemCache() {}
+
+#endif
+
+}
+}
--- a/backends/implot_impl_opengl3.cpp
+++ b/backends/implot_impl_opengl3.cpp
@@ -43,6 +43,9 @@ struct HeatmapShader
 	GLuint AttribLocationProjection = 0; ///< Attribute location for the projection matrix uniform
 	GLuint AttribLocationMinValue = 0;   ///< Attribute location for the minimum value uniform
 	GLuint AttribLocationMaxValue = 0;   ///< Attribute location for the maximum value uniform
+	GLuint AttribLocationAxisLog = 0;    ///< Attribute location for the logarithmic axes uniform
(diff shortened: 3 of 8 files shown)


marcizhu:
I've run a small example I showcased some time ago in #180 ([here](https://github.com/epezent/implot/discussions/180#discussioncomment-416276)), which consisted of a heatmap with dimensions 1024x768. Back then I got 60-80 FPS consistently, and now...

image

I get some peaks of even 300 FPS but since the time of each frame is so small (~3ms) even changes of a few microseconds can make the FPS counter go up or down by huge amounts. CPU usage is about 20% (which seems OK to me, since that waveform is being generated each frame, together with the random noise) and GPU usage is about 80-85%, so this means my code might have some bottlenecks somewhere in there. Still, having ~280 FPS consistently on a MacBook Pro 2018 is not bad at all!

epezent:
@marcizhu, first of all this is freakin' awesome! Thanks for taking the lead on this. The results really speak for themselves. My little spectrogram demo below runs almost 3.5x faster. I'm sure the gains are even more extreme for higher resolution heatmaps -- I'll have to test a bit more later. 

## Feedback to Comments

> 1. OpenGL doesn't support `double` nor `*int64_t` (aka `ImU64`/`ImS64`). If this kind of data is passed to a heatmap, it is internally converted to `float` or `ImU32`/`ImS32` (respectively).

This seems fine, and is probably the best we can do. However, I noticed that each `OpenGL3_SetHeatmapData` that needs to convert data is making a (potentially very large) dynamic allocation every frame. I would recommend making an `ImVector<float>` member  in `OpenGLContextData` and reusing this throughout the lifetime of the program, resizing as needed.

> 2. When using OpenGL acceleration, the interpolation tables for heatmaps are unnecessary. Maybe we could only enable them in case this optimization is disabled

These tables are used for things other than heatmaps. We need to keep them around for now.

> 3. Right now I use the same trick ImGui's `main.cpp` file uses to access OpenGL functions, which is rely on a macro definition to include one of the many supported libraries. It is possible to implement our own OpenGL loader so that we don't rely on any external library (ImGui will still depend on that library though), but that would add a bit of complexity maybe.

Sounds good to me. Note that `imgui_impl_opengl3.h` has some extra header magic in it so you don't have to manually define the loader:

https://github.com/ocornut/imgui/blob/7c44d067e8d1662549db73f80099e01a9c1a5f46/backends/imgui_impl_opengl3.h#L67-L86


> 4. Log-Log, Lin-Log and Log-Lin axis are not yet implemented.

Understandable. Should we fallback to the software rasterizer for now, or do you see an easy path forward for OpenGL support? If not, what is primary issue to overcome?

> I tried to implement this as straightforward as possible, and with as little modifications to the 'core' of the library as possible. Also, all the implementation details are inside the file `backends/implot_opengl.cpp`. If the macro `IMPLOT_ENABLE_OPENGL3_ACCELERATION` is defined, then that file will compile and the library will use the shader-accelerated code. If not, it will default to the current method.

I like the file organization, as it mimics the ImGui repo. I think we may have to change macro names and logic a bit for when/if we ever support other backends. If we stick to this convention, then we'll eventually have `IMPLOT_ENABLE_DX11_ACCELERATION`, `IMPLOT_ENABLE_METAL_ACCELERATION`, etc. all over the place. Let's think more on it and see if there a clever way of handling it. 

## General Feedback

- `OpenGLContextData` is currently a static global. I think it would be better to make this owned by the `ImPlotContext` if possible. I know it makes the includes and #ifdefs a bit more complicated, but ImPlot and ImGui generally avoid static global in the source files. 
- We should probably put initialization code in `ImPlot::CreateContext`, or have it called from there somehow. For example, you currently use lazy initialization for the shader program, which is fine, but it might be cleaner to put it in `CreateContext`.

- Oh, and **everything is rendered upside down**  :P

image


ImPlot Default
image

ImPlot OpenGL
image

I will continue to review this over the next few days. Since this PR is our first step in what I hope to be a more comprehensive effort to GPU accelerate ImPlot, I want to take things slow  and ensure we are going to set ourselves up for success in the future. I may ask for more changes, or make some edits myself before merging.

Thanks again, Marc! This is really cool and takes ImPlot to the next level!

marcizhu:
Thanks for the feedback! Let me reply to some of those comments:

> This seems fine, and is probably the best we can do. However, I noticed that each `OpenGL3_SetHeatmapData` that needs to convert data is making a (potentially very large) dynamic allocation every frame. I would recommend making an `ImVector<float>` member in `OpenGLContextData` and reusing this throughout the lifetime of the program, resizing as needed.

I agree, we should have some sort of storage so that we don't `new`/`delete[]` on each frame. To be honest those three functions were just a quick and dirty way to test if the library worked and I forgot about them :P
I will add an `ImVector` member to the context data for this kind of things. Thanks!

> Understandable. Should we fallback to the software rasterizer for now, or do you see an easy path forward for OpenGL support? If not, what is primary issue to overcome?

I think it's not super hard to support, so this should be doable in the near future. We just need to pass some uniforms to the shader and that's it :)

> I like the file organization, as it mimics the ImGui repo. I think we may have to change macro names and logic a bit for when/if we ever support other backends. If we stick to this convention, then we'll eventually have `IMPLOT_ENABLE_DX11_ACCELERATION`, `IMPLOT_ENABLE_METAL_ACCELERATION`, etc. all over the place. Let's think more on it and see if there a clever way of handling it.

This got me thinking. I agree that once we start having more backends there will be tons of `#ifdef`/`#endif` all over the place (that's why I tried to modify the library as little as possible, to prevent this). I think we can do something about this: what if we define a standard set of functions that backends should support, and depending on which macro is defined the _implementation_ of those functions changes. For example, in a header file we could declare the functions `Backend::SetHeatmapData()`, `Backend::RenderHeatmap()`, `Backend::AddColormap()`, etc... If no macro is defined, then those functions will be a noop (which removes the need of the `#ifdef`/`#endif` in most cases). If for example the macro `IMPLOT_ENABLE_OPENGL_ACCELERATION` is defined, then the file `implot_opengl.cpp` will be compiled and those functions will implement the OpenGL acceleration. If instead `IMPLOT_ENABLE_DX11_ACCELERATION` was defined then the file `implot_directx11.cpp` would be compiled and those functions would contain the DirectX 11 implementation.

What do you think about this? I think it's one of the best things we can do to remove `#ifdef`s from the codebase and pave the way for future backends.

> `OpenGLContextData` is currently a static global. I think it would be better to make this owned by the `ImPlotContext` if possible. I know it makes the includes and #ifdefs a bit more complicated, but ImPlot and ImGui generally avoid static global in the source files.

To be honest that was my first idea, but I thought that maybe you wanted to 'divide' ImPlot's context from this OpenGL 'context' data. Having an `OpenGLContextData*` inside ImPlot's context is the way to go I think. If we do the changes I explained in the previous comment this should be a pretty easy one :)

> We should probably put initialization code in `ImPlot::CreateContext`, or have it called from there somehow. For example, you currently use lazy initialization for the shader program, which is fine, but it might be cleaner to put it in CreateContext.

There is no other way to initialize the shader. Maybe I should have explained this before. Here's the issue: this shader has to be layout-compatible with ImGui's shader so that it receives the vertex, UV and color data that ImGui sends to the renderer (in other worlds, by just changing one shader for the other everything should work). The issue is that to do this I need to have access to either the shader attribute locations (which are private variables inside `imgui_impl_opengl3.cpp`) or I need access to the shader handle (which is also a private variable there). ImGui doesn't bind the shader either so `glGetIntegerv(GL_CURRENT_PROGRAM, (GLint*)&CurrentShader)` doesn't work either. The only way to get access to the shader is mid-rendering, because at that point the shader **is** bound and `glGetIntegerv(GL_CURRENT_PROGRAM, (GLint*)&CurrentShader)` _does_ work. That's why I did this weird lazy initialization for the shader :/

If you have a better way to handle this, I'm fully open to changes. To be honest I dislike this kind of initialization but It's the best I have so far...

> Oh, and **everything is rendered upside down** :P

Woops! My bad! I forgot OpenGL renders everything from the bottom-left corner. This is fixed and the next commit will solve this.

> I will continue to review this over the next few days. Since this PR is our first step in what I hope to be a more comprehensive effort to GPU accelerate ImPlot, I want to take things slow and ensure we are going to set ourselves up for success in the future. I may ask for more changes, or make some edits myself before merging.

Sure! You can ask for as many changes as you want! Yesterday I made this PR rather quickly and it's really rough and more of a proof-of-concept rather than a final product.

epezent:
> I think it's not super hard to support, so this should be doable in the near future. We just need to pass some uniforms to the shader and that's it :)

Cool! Let's support this before merging if possible.

> This got me thinking. I agree that once we start having more backends there will be tons of #ifdef/#endif all over the place (that's why I tried to modify the library as little as possible, to prevent this). I think we can do something about this: what if we define a standard set of functions that backends should support, and depending on which macro is defined the implementation of those functions changes. For example, in a header file we could declare the functions Backend::SetHeatmapData(), Backend::RenderHeatmap(), Backend::AddColormap(), etc... If no macro is defined, then those functions will be a noop (which removes the need of the #ifdef/#endif in most cases). If for example the macro IMPLOT_ENABLE_OPENGL_ACCELERATION is defined, then the file implot_opengl.cpp will be compiled and those functions will implement the OpenGL acceleration. If instead IMPLOT_ENABLE_DX11_ACCELERATION was defined then the file implot_directx11.cpp would be compiled and those functions would contain the DirectX 11 implementation.

This plan sounds good to me. The only issues that I can fathom are if one backend needs more arguments in the function than another (not sure why, just postulating), or if one backend can't, for whatever reason, implement all the backend functions. We can cross that bridge when we get there, though. 

One idea worth considering is putting the backend implementation in a `.hpp` file, and then in `implot_items.cpp`:

```cpp
#ifdef  IMPLOT_ENABLE_OPENGL3_ACCELERATION 
    #include "backends/implot_opengl.hpp`
#elif  IMPLOT_ENABLE_DX11_ACCELERATION 
    #include "backends/implot_dx11.hpp
...
#endif
```

We don't have to, but it saves the extra step of adding `implot_opengl.cpp` to your sources. 

> To be honest that was my first idea, but I thought that maybe you wanted to 'divide' ImPlot's context from this OpenGL 'context' data. Having an OpenGLContextData* inside ImPlot's context is the way to go I think. If we do the changes I explained in the previous comment this should be a pretty easy one :)

Yea, possibly we have a `void* Backend::CreateContext()` and a `Backend::DestroyContext(void*)` function, and let the `void*` be owned by the `ImPlotContext`. In this case, the `void*` would be `OpenGLContextData*`

> There is no other way to initialize the shader. Maybe I should have explained this before. Here's the issue: this shader has to be layout-compatible with ImGui's shader so that it receives the vertex, UV and color data that ImGui sends to the renderer (in other worlds, by just changing one shader for the other everything should work).

I don't have a complete understanding, so maybe this is a stupid question, but how are we going to handle cases where the user is not using ImGui's default OpenGL backend, and instead is using their own? Is there any possibility that their shader would have a different layout? 

Otherwise, I trust your judgment here. I'm not very knowledgeable of OpenGL. 

> Woops! My bad! I forgot OpenGL renders everything from the bottom-left corner. This is fixed and the next commit will solve this.

Sweet! It looks perfect now:

image

epezent:
One other thing that I'm excited about is the possibility of changing `GL_NEAREST` to `GL_LINEAR` inside of `CreateHeatmapTexture`. I just played around with this and it makes the spectrogram very smooth. Is there a clean way to expose this option to the end user?

marcizhu:
> This plan sounds good to me. The only issues that I can fathom are if one backend needs more arguments in the function than another (not sure why, just postulating), or if one backend can't, for whatever reason, implement all the backend functions. We can cross that bridge when we get there, though.

Yeah, that's something that worries me too, but I guess if that ever happens and a function `fn()` needs parameters `x`, `y`, `z` on OpenGL and `x`, `y`, `w` on DirectX, we could just have `fn` be something like `fn(x, y, z, w)` for both. OpenGL would ignore `w` and DirectX would ignore `z`.

> One idea worth considering is putting the backend implementation in a .hpp file [...]

I agree, we should consider this too. The main downside to this is that compilation times might be a bit higher, and we still have the issue of having lots of `#ifdef`/`#endif` all over the place, but it would help users since there would be no additional C++ files to compile.

To be honest, I feel like having a common header for all backends and a `.cpp` file for each backend is the way to go about this. It would remove most of the `#ifdef`s currently on the code and provide a "standard" interface for future backends. But it's _your_ library after all!

> I don't have a complete understanding, so maybe this is a stupid question, but how are we going to handle cases where the user is not using ImGui's default OpenGL backend, and instead is using their own? Is there any possibility that their shader would have a different layout?

Good question. As far as I know, if their shader/backend is compatible with ImGui it needs a position attribute, a UV attribute and a color attribute. My shader only uses the position & UV attribute, so it should be compatible with any other shader out there. The only issue might be if they use other names for those attributes, in which case it's just a matter of changing the names on our shader too. Maybe those shader attributes should be defined in a `#define` or constant for easier changes?

> One other thing that I'm excited about is the possibility of changing `GL_NEAREST` to `GL_LINEAR` inside of `CreateHeatmapTexture`. I just played around with this and it makes the spectrogram very smooth. Is there a clean way to expose this option to the end user?

Yeah, that's definitely an option! I didn't implement this because the original code didn't provide this option, but I guess it's just a matter of adding a `bool` on `PlotHeatmap()` and forwarding this to the OpenGL backend. I can do it if you want :)

epezent:
> To be honest, I feel like having a common header for all backends and a .cpp file for each backend is the way to go about this. It would remove most of the #ifdefs currently on the code and provide a "standard" interface for future backends. But it's your library after all!

I agree. This seems like the most logical approach.

> Good question. As far as I know, if their shader/backend is compatible with ImGui it needs a position attribute, a UV attribute and a color attribute. My shader only uses the position & UV attribute, so it should be compatible with any other shader out there. The only issue might be if they use other names for those attributes, in which case it's just a matter of changing the names on our shader too. Maybe those shader attributes should be defined in a #define or constant for easier changes?

Hmm, I'm not familiar enough with the process to make a call right now. Let me educate myself on the topic a bit more, and dig through the ImGui backend itself. 

**IF THERE ARE ANY RESIDENT OPENGL EXPERTS LURKING, PLEASE CHIME IN!**

> Yeah, that's definitely an option! I didn't implement this because the original code didn't provide this option, but I guess it's just a matter of adding a bool on PlotHeatmap() and forwarding this to the OpenGL backend. I can do it if you want :)

I've been contemplating adding optional flags for certain plot types, e.g. `ImPlotHeatmapFlags_XXX`. If we do that, then we could do something like what ImGui does in its internal API:

`implot.h`
```cpp
enum ImPlotHeatmapFlags_ {
    ImPlotHeatmapFlags_None = 0,
    ImPlotHeatmapFlags_ColMajor = 1 << 0,
    ...
    ImPlotHeatmapFlags_N = 1 << n
}
```

`implot_gpu.h` (our *common* header file, call it whatever)

```cpp
ImPlotHeatmapFlagsGPU_ {
    ImPlotHeatmapFlags_Linear = 1 << n+1
    ...
}
```

So, the user gets a few extra flags if and only if they have a GPU backend enabled. We can have `implot.h` include `implot_gpu.h` automagically somehow.

marcizhu:
> I agree. This seems like the most logical approach.

Alright, I'm gonna do this change. I like the name `implot_gpu.h`.

> Hmm, I'm not familiar enough with the process to make a call right now. Let me educate myself on the topic a bit more, and dig through the ImGui backend itself.

About this, if the user changes the shader and/or backend, the most they will have to do is change this strings in lines 74 and 75 of file `implot_opengl3.cpp`:

https://github.com/marcizhu/implot/blob/aa716726da99bde5cb6af16a38e5af104c18eb14/backends/implot_opengl3.cpp#L74-L75

The idea of adding heatmap flags seems good to me and easy to implement. It's just as easy as adding a new parameter to `Backend::RenderHeatmap()`. For now though, I will implement this common header, remove all of the `#ifdef`s from the code and clean everything up a bit. Let me know if you want me to do anything else! :D

marcizhu:
@epezent I discovered that integer heatmaps weren't working as expected; only `float` heatmaps worked correctly. This is now fixed as of the last commit. Now I use two shaders (OpenGL doesn't like mixing float sampler2D and int sampler2D textures in the same shader...). This is a bit of an inconvenience but it forced me to rewrite the code such that it is quite easy to support multiple shaders. I also took this opportunity to clean the code a bit and fix some other issues, like the shaders not being deleted on `DestroyContext()`.

Anyway, now both all heatmaps should work properly (the ones on the left use an array of `float`, the one on the right uses an `int` array):
image
Tomorrow I will try to implement the log-transformations for both axes. Wish me luck! :D

epezent:
A few recommendations and things I noticed:

- For consistency, I actually prefer the name `implot_backend.h` over `implot_gpu.h`. 
- I think we should pass `gp.CurrentItem->ID`  instead of `gp.CurrentPlot->ID` to the backend functions in the case where two heatmaps are on one plot.
- I'm wondering if `SetHeatmapData` is generic enough that it should just be exposed as `SetTextureData2D` or something like that.
- Thinking ahead, I propose that we handle `#ifdef`'s in the following way:

1) Change `IMPLOT_ENABLE_OPENGL3_ACCELERATION` to `IMPLOT_BACKEND_ENABLE_OPENGL3` (for consistency)

2) In `implot_backend.h`, add something like this:

```cpp
#ifdef IMPLOT_BACKEND_ENABLE_OPENGL3
    #define IMPLOT_BACKEND_ENABLED
    #define IMPLOT_BACKEND_HAS_HEATMAP
    #define IMPLOT_BACKEND_HAS_XXXXXXX
#elif IMPLOT_BACKEND_ENABLE_DX11
    #define IMPLOT_BACKEND_ENABLED
    ...
#endif
```

3) Then, replace `IMPLOT_ENABLE_OPENGL3_ACCELERATION` in `implot_items.h` with `IMPLOT_BACKEND_HAS_HEATMAP`. 
4) We will add to this as more plot types and backends are supported. This saves us having to do nasty things like `#ifdef IMPLOT_BACKEND_ENABLE_OPENGL3 | IMPLOT_BACKEND_ENABLE_DX11` in the main source files. 

Once you have log-scale working, I'll make another thorough pass through and clean up a few nitpicky things. Beyond that, I can't think of anything else we need to do right now. I suspect this will evolve over time as we add more plot types. Hopefully I have some time to investigate GPU accelerated line triangulation in the near future.

marcizhu:
I love where this is going. Very clever! I did all this changes (except renaming `SetHeatmapData()` to `SetTextureData2D()` which I can do in no time if you want) in the latest commit.

If there's any other issue or I missed something, just let me know and I'll do it as soon as possible 😄 
Thanks for all this great feedback!

bear24rw:
Really excited about this feature! I would be interested in adding a metal backend since that's my primary backend but there are a few complications in doing that with the current architecture:
- The backend needs to be passed the `MTLDevice` to initialize the shaders / pipeline
- Every frame needs to be passed the `MTLCommandBuffer` and `MTLRenderCommandEncoder`

These all come from outside of imgui (imgui metal backend prototypes [here](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_metal.h)).

I'm wondering if an alternative to the current setup would be to more directly model imgui and create `backends/implot_impl_opengl3.{cpp,h}`, `backends/implot_impl_metal.{cpp,h}`, etc. The user would be required to call `ImPlot_ImplOpenGL3_Init `/`ImPlot_ImplMetal_Init(device)`, same as imgui.

Before each frame you can setup the implot backend with the current buffer and encoder for the frame, something like:
```cpp
// init
ImGui::CreateContext();
ImPlot::CreateContext();
ImGui_ImplMetal_Init(device);
ImPlot_ImplMetal_Init(device);

// render loop
ImGui::Render();
ImPlot_ImplMetal_RenderSetup(command_buffer, render_encoder);
ImGui_ImplMetal_RenderDrawData(ImGui::GetDrawData(), command_buffer, render_encoder);
```

I'm not familiar with the other imgui backends (like dx12 / vulkan) but they might have the same issue as the metal one. OpenGL is a little unique in that is has global state.

epezent:
@bear24rw , thanks for this insight! I'm glad we are having these discussions now.

I took a look at the other ImGui backends, and this seems to be a common pattern. We should probably have a `ImPlot_ImplMetal_Init` and `ImPlot_ImplMetal_RenderSetup` for each backend. At first I was going to suggest we could keep these in `implot_backends.h`, wrapped in the `#ifdef`s above. But looking a little more closely, it seems like some require struct forward declarations and other miscellanea, so probably best to have a `backends/implot_impl_metal.{cpp,h}` for each backend.

Glossing over details such as naming convention, namespaces, etc., my current thinking is:

`implot_backends.h`

```cpp
#ifdef IMPLOT_BACKEND_ENABLE_METAL
    #include "implot_impl_metal.h"
#ifdef IMPLOT_BACKEND_ENABLE_OPENGL3
    #include "implot_impl_opengl3.h"
#elif IMPLOT_BACKEND_ENABLE_DX11
    #include "implot_impl_dx11.h"
#endif

#ifdef IMPLOT_BACKEND_ENABLED
  void* CreateContext();
  void DestroyContext();
  void BustPlotCache();
  void BustItemCache()
#endif

#ifdef IMPLOT_BACKEND_HAS_COLORMAP
  void AddColormap(const ImU32* keys, int count, bool qual);
#endif

#ifdef IMPLOT_BACKEND_HAS_HEATMAP
  void SetHeatmapData(int plotID, const T* values, int rows, int cols);
#endif
...
```

`implot_impl_metal.h`, `implot_impl_opengl3.h`, `implot_impl_dx11.h`, etc.
```cpp
#include "implot_backend.h"     

#define IMPLOT_BACKEND_ENABLED
#define IMPLOT_BACKEND_HAS_HEATMAP
#define IMPLOT_BACKEND_HAS_COLORMAP
#define IMPLOT_BACKEND_HAS_XXXXXXX

@class MTLRenderPassDescriptor;
@protocol MTLDevice, MTLCommandBuffer, MTLRenderCommandEncoder;

bool ImPlot_ImplMetal_Init(id<MTLDevice> device);
void ImPlot_ImplMetal_RenderDrawData(id<MTLCommandBuffer> commandBuffer, id<MTLRenderCommandEncoder> commandEncoder);
```


`implot_impl_metal.cpp`, `implot_impl_opengl3.cpp`, `implot_impl_dx11.cpp`, etc.

```cpp
#include "implot_implt_metal.h"

bool ImPlot_ImplMetal_Init(id<MTLDevice> device) { ... }
void ImPlot_ImplMetal_RenderDrawData(id<MTLCommandBuffer> commandBuffer, id<MTLRenderCommandEncoder> commandEncoder)  { ... }
void* CreateContext() { ... }
void DestroyContext()  { ... }
void BustPlotCache()  { ... }
void BustItemCache()  { ... }
void AddColormap(const ImU32* keys, int count, bool qual)  { ... }
void SetHeatmapData(int plotID, const T* values, int rows, int cols)  { ... }
```

There's probably a cleaner solution I'm no considering. Open to any suggestions, so let me know your thoughts. 

@bear24rw, are you comfortable taking a stab at a Metal implementation of @marcizhu's heatmap renderer? I think having two examples working will clear things up even more.

marcizhu:
@epezent I've added this changes. I didn't add `ImPlot_ImplOpenGL_Init()` nor `ImPlot_ImplOpenGL_RenderDrawData()` since for OpenGL both would be empty functions. We can add them later if we want a more 'standard' interface for all backends.

@bear24rw if you start working on a metal implementation for heatmaps feel free to ask me anything. I can't help you much about Metal but I can explain you how the OpenGL implementation works and hopefully that will give you some ideas on how to implement the same features on Metal.

marcizhu:
@epezent logarithmic X & Y axes are now implemented! I haven't done a lot of testing, but so far it looks like it works fine. If there's any issue or anything you want me to implement, just let me know! 😄 

image

epezent:
Sweet! I'll check it out later. One thing I kind of want is a runtime bool to enable/disable GPU acceleration if it's enabled. Probably in ImPlotStyle. Default to false unless IMPLOT_BACKEND_ENABLED defined, then default to true. This will let us test things more quickly and offers users the ability to selectively accelerate plots.

epezent:
I added a quick and dirty runtime switch and a few issues became apparent. 


1) The GPU color doesn't seem to be sampling very linearly. My software implementation may not be perfect either, but this seems off.

ex2

2) Not sure what's going on here, but something's causing the image to shift (you can see some more color crushing here too).

ex1


3) LogX looks good...

ex3

... but LogY doesn't:

ex4

bear24rw:
> There's probably a cleaner solution I'm no considering. Open to any suggestions, so let me know your thoughts.

This seems reasonable to me at first glance.
 
> @bear24rw, are you comfortable taking a stab at a Metal implementation of @marcizhu's heatmap renderer? I think having two examples working will clear things up even more.

I can certainly take a look at implementing it but my schedule for the next month or two is really limited so no promises. If there are any other metal users lurking feel free to take a first stab!

marcizhu:
@epezent I'll take a look at this issues but I think I have an idea of what's going on:

1. OpenGL by default allows wrapping textures, so you can sample a texture from 0 to 1, and if you give it a negative number or something greater than 1, it just wraps around. We don't want that on colormaps, so I set it to `GL_CLAMP`. This means now that everything smaller than 0 is white (using that colormap) and everything greater than 1 is black. This affects a bit the linear interpolation making it a bit more agressive towards the extremes (basically when you're near 0 or near 1, it tends towards bright white or dark black). I think that's the issue here, I'll do some research just to see if I can set OpenGL to don't wrap textures at all.

2. This one looks like it's vertically flipped for some reason. Can you give me some source to test this? Maybe you could commit the changes you did and we can continue from there.

3. The LogY does this because the heatmp starts at 0 and it basically stretches it to -inf. Setting the minimum Y position to something like 0.001 does the trick and works fine. The older implementation didn't have this issue since it rendered a bunch of small quads _from the center point_, and thus when it rendered the bottom one, the Y coordinate (or X for that matter) was always < 0.0 and was treated accordingly. In the GPU implementation this doesn't happen since I'm rendering the entire quad at once and so the point `(X,0)` goes to infinity. I think I can patch that one easily so that both heatmaps work the same and there's no need to change anything.

epezent:
1) That's quite unfortunate. We definitely need to find an acceptable solution to this. One idea, as you mentioned, is to pass more keys. Actually, we already have that in the colormap interpolation table we compute on the CPU. If you use this to form the colormap texture, you might even be able to to turn GL_LINEAR off and use GL_NEAREST, which is basically what the CPU does by indexing this table.

image

Manual computation sounds interesting, but I don't understand the details. Can you elaborate on this comment:

> maybe manually compute the colors on the GPU (but that becomes an issue when the user wants to use their own colormaps)

2. Ah, totally forgot we had that bool in the implementation. Good catch and fix.
3. Makes perfect sense. I suppose we could clamp 0 to something like FLT_EPSILON to prevent this.

marcizhu:
> Can you elaborate on this comment: maybe manually compute the colors on the GPU (but that becomes an issue when the user wants to use their own colormaps)

Sure thing! Basically now I send the colors to the GPU as a 1D texture. Then, using the value on the heatmap 2D texture I sample that colormap and get a color, which is the color of that pixel. An alternative would be to ditch the idea of this 1D texture and instead implement colormaps as a function that takes a `float` from `0.0` to `1.0` and outputs a color (see [kbinani/colormap-shaders](https://github.com/kbinani/colormap-shaders); for example [this shader](https://github.com/kbinani/colormap-shaders/blob/7ddaa34978c4729cfe6eddaf7bf68bb4831760bb/shaders/glsl/MATLAB_jet.frag) implements the MATLAB "Jet" colormap using a function inside the shader). This solution might be slightly viable but it becomes a problem once we start considering that users might supply their own colormaps, and we shouldn't have to ask them for shaders or anything like that, so really this solution, while it would certainly fix the issue for the built-in colormaps (you can see this on the pic I sent), still wouldn't solve this problem for user-added colormaps.

I think our best option here is to store more colors. Maybe we don't need the entire `255*(k-1)+1` array of colors, but a 10x oversample might give nice enough results to be pretty much indistinguishable with the naked eye. I'll do some experiments and let you know how it goes :smile:

epezent:
Ok, I understand now. Yea, that's probably not the direction we want to go. Let's see how far oversampling gets us.

epezent:
@marcizhu , checkout `ImDrawCallback_ResetRenderState` in imgui.h and imgui_impl_opengl3.cpp. I think you can eliminate your `ResetState` callback with this.

marcizhu:
I tried using `ImDrawCallback_ResetRenderState` back when I started this PR, but it doesn't unbind the 1D texture used for colormaps and so it makes other windows glitch 😞 
Using my custom function to unbind it was the only solution that I found.

Thanks for the suggestion though 👍🏻

epezent:
Been playing around with this and getting some very impressive results. Here's a 1000x1000 heatmap where there's a 50x performance increase!

image


image

marcizhu:
Woah! That's impressive!! I guess that the bigger the heatmap, the more performance gain there is (as you said).
Before opening this PR I created a 1000x1000 heatmap just to test if it worked at around ~60 and indeed it capped to 60 FPS due to VSync, but I never tried that same code with VSync disabled. Seeing this 50x increase is pretty awesome :D

PS: The colormap thingy might get delayed a few days; sadly I've been having some issues with my laptop and I might need a couple of days to fix them. I hope to have a definitive answer (and hopefully a commit that fixes it) in 3-4 days.

epezent:
> PS: The colormap thingy might get delayed a few days; sadly I've been having some issues with my laptop and I might need a couple of days to fix them. I hope to have a definitive answer (and hopefully a commit that fixes it) in 3-4 days.

Sounds good. We're not in a rush. Also, I think the best way to roll this out at first will be on a separate branch similar to ImGui `docking`. Seems a bit premature to put this in `master` with only one backend and one plot type. At the very least I'd also like to have GPU accelerated lines working before making this "public". I've started some experimentation with this.

marcizhu:
Sounds good to me :D
I think I can retarget this PR at any time, so we can change the branch just before merging (and I'm pretty sure you can do this too).

If you need help implementing GPU-accelerated lines (or any other feature) just let me know and I'll help you where I can. I'm not an expert in OpenGL by any means but you know, maybe I can help you implementing some features or figuring some things out 😉

marcizhu:
I'm back! Sorry for the long delay, I had to send my Macbook for a repair and I got it back today. Anyway, I got back to working and here's what my latest commit does:
image

Pretty amazing huh? Now CPU and GPU implementations should look identical to one another. I might be able to get this back to how it was before (use the provided colormap keys instead of manually interpolating them and sending those to the GPU, thus reducing GPU memory usage) but at least now we don't have that skew towards the extremes of the colormap :D

If there's any other issues that have arised in these days or anything else you want me to do/fix, just let me know as always!
Sorry once again for the delay 😞

marcizhu:
Okay, that was quick. It turns out that OpenGL interpolates using the _center_ of the pixels. That means that for example in the case of the "greys" colormap, `0.25` (the center of the first pixel) would be pure white and `0.75` would be pure black, and anything between those values would be grey. But that also means that anything _under_ `0.25` would also be white and anything over `0.75` would also be black, thus skewing the colors.

I force-pushed my latest commit so that I don't pollute this PR with useless commits. As of the latest commit, we still submit the colormap keys to the GPU as before but now heatmap values are not interpolated to values between `0.0` and `1.0` but instead to the center of the first pixel and up to the center of the last pixel. This should be most noticeable on smaller colormaps (like the "greys" colormap).

You can ignore my last comment. And sorry for the spam :P

bear24rw:
> > There's probably a cleaner solution I'm no considering. Open to any suggestions, so let me know your thoughts.
> 
> This seems reasonable to me at first glance.

I think there is an issue with this as it will bring platform specific things into the platform agnostic cpp code. For instance, the metal impl header needs the objcpp lines:
```objc
@class MTLRenderPassDescriptor;
@protocol MTLDevice, MTLCommandBuffer, MTLRenderCommandEncoder;
```
but those will get pulled into `implot_backend.h` which gets pulled into `implot_items.cpp` resulting in syntax errors when compiling the cpp file.

marcizhu:
> but those will get pulled into `implot_backend.h` which gets pulled into `implot_items.cpp` resulting in syntax errors when compiling the cpp file.

Why can't you just add those lines in the .cpp file? An `#include` will just replace that line with the contents of the specified file. I think it's easy to just add those two lines inside `impot_impl_metal.m`. That way there won't be syntax errors on `implot_items.cpp`. Am I missing something?

bear24rw:
> > > There's probably a cleaner solution I'm no considering. Open to any suggestions, so let me know your thoughts.
> > 
> > 
> > This seems reasonable to me at first glance.
> 
> I think there is an issue with this as it will bring platform specific things into the platform agnostic cpp code. For instance, the metal impl header needs the objcpp lines:
> 
> ```objc
> @class MTLRenderPassDescriptor;
> @protocol MTLDevice, MTLCommandBuffer, MTLRenderCommandEncoder;
> ```
> 
> but those will get pulled into `implot_backend.h` which gets pulled into `implot_items.cpp` resulting in syntax errors when compiling the cpp file.

A potential, but maybe not the best, solution is to maintain each backends capabilities in `implot_backend.h`. So instead of including the `impl.h` files depending on which backend is enabled, we would specify `IMPLOT_BACKEND_HAS_COLORMAP`, etc for each one. Then the `impl.h` files include `implot_backend.h` to get the function prototypes that the core implot library expects. They can also add their implementation specific prototypes, like `ImPlot_ImplMetal_RenderSetup(...);`. The end user should never include `implot_backend.h` in their `main.cpp`, they should always include the `impl.h` for the render backend they are using. I think this will solve the issue of objc syntax leaking into the implot cpp files. Let me know if this doesn't make sense and I can provide a more concrete example.

bear24rw:
> > but those will get pulled into `implot_backend.h` which gets pulled into `implot_items.cpp` resulting in syntax errors when compiling the cpp file.
> 
> Why can't you just add those lines in the .cpp file? An `#include` will just replace that line with the contents of the specified file. I think it's easy to just add those two lines inside `impot_impl_metal.m`. That way there won't be syntax errors on `implot_items.cpp`. Am I missing something?

The metal (and others) backend needs to be passed a `MTLCommandBuffer` and `MTLRenderCommandEncoder` every frame, so they are parameters of the public facing interface. Its the same situation that imgui has:

https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_metal.h#L21

The thing that is different about implot is that implot is trying to call backend methods, so it needs to include their headers. imgui does not call the backend methods so it doesn't hit this same issues.

So I am suggesting we have a `implot_backend.h` which implot can include, and the backends can include, then the _actual_ backend specific stuff goes in the `impl.h` files which implot never looks at.

epezent:
@marcizhu, the new colormap solution works fantastic! I can't tell any appreciable difference in the colors now. If anything, the the GPU version looks slightly sharper, which is a good thing! If you give me permissions, I can push my change that allows for runtime switching between software and hardware rendering.

> A potential, but maybe not the best, solution is to maintain each backends capabilities in `implot_backend.h`. So instead of including the `impl.h` files depending on which backend is enabled, we would specify `IMPLOT_BACKEND_HAS_COLORMAP`, etc for each one. Then the `impl.cpp` files include `implot_backend.h` to get the function prototypes that the core implot library expects. They can also add their implementation specific prototypes, like `ImPlot_ImplMetal_RenderSetup(...);`. The end user should never include `implot_backend.h` in their `main.{cpp,mm}`, they should always include the `impl.h` for the render backend they are using. I think this will solve the issue of objc syntax leaking into the implot cpp files. Let me know if this doesn't make sense and I can provide a more concrete example.

@bear24rw , makes sense to me! I don't think it's too chaotic to have to put the feature defines in `implot_backend.h` away from `impl.h`. I'm still convinced there's a cleaner solution, but I haven't thought of it yet. 

PS: I've been experimenting with line rendering. Initial tests with a 1,000,000 point curve shows performance of ~400-600 FPS vs 15 FPS using the current rendering method. I've tested a few different approaches, and still have more to try. Quite excited!

bear24rw:
@epezent, can we get @marcizhu stuff merged into a branch to make it easier to experiment with some reorg/other backend PRs?

epezent:
> @epezent, can we get @marcizhu stuff merged into a branch to make it easier to experiment with some reorg/other backend PRs?

That would be logical. Yes, I can do this :)

epezent:
Pushed to `backends` branch

epezent:
Let's move our discussion to #261

marcizhu:
That was quicker than I expected! Thanks for merging, @epezent!!

tomgwasira:
Hello. Thank you for implementing this feature. I would like to find out if there are any usage instructions esp. with regards to dependencies. I am struggling to get this branch running out the box as there are dependencies that are not being included. For instance, in `implot_impl_opengl3.cpp` I get errors such as `GL_TEXTURE_2D is undefined`. I am using GLFW. Thanks.

PR #264, opened 2021-07-09
fix compile issue with official imgui

toxieainc:


epezent:
Fixed in a separate commit. Thanks let me know quickly. Totally missed this.

PR #269, opened 2021-07-29
PlotShaded creates Opposite triangle winding

jeslaspravin:
Triangle winding in this line creates one triangle clockwise and another counter-clockwise. This leads to one of the triangles being culled when culling is enabled.

Clockwise
https://github.com/epezent/implot/blob/507459fd5f7d6b210ad2a8663d0ae765251fcedf/implot_items.cpp#L672-L674

Counter-clockwise
https://github.com/epezent/implot/blob/507459fd5f7d6b210ad2a8663d0ae765251fcedf/implot_items.cpp#L675-L677


epezent:
Good catch and thanks for the fix!

PR #280, opened 2021-08-26
added #include<xmmintrin.h> directive for sse when needed

unevens:
hi, thanks for this great library! 
i needed to add that header to make it build on macOS. 
cheers.


epezent:
Can you see if https://github.com/epezent/implot/commit/5ed9e786761958a8cb8f48888108ce6ae6818683 fixes the issue for you. I was assuming the header would trick down from `imgui_internal.h`, but your version of ImGui may predate that addition.

unevens:
it works, thank you

PR #322, opened 2022-02-09
Fix build failure on pre C++11 compilers

RT222:
The latest version of implot cannot build on pre C++11 compilers because it uses delegated constructors and consecutive right angle brackets with no space between them. I don't know if targeting pre C++11 compilers is in implot goals, but if it is, this PR will fix that.

### DIFF
--- a/implot_items.cpp
+++ b/implot_items.cpp
@@ -445,13 +445,15 @@ struct GetterError {
 // Transforms convert points in plot space (i.e. ImPlotPoint) to pixel space (i.e. ImVec2)
 
 struct TransformerLin {
-    TransformerLin(double pixMin, double pltMin, double,       double m, double    ) : PixMin(pixMin), PltMin(pltMin), M(m) { }
+    TransformerLin(double pixMin, double pltMin, double, double m, double) : PixMin(pixMin), PltMin(pltMin), M(m) { }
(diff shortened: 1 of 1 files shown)


epezent:
@RT222 , I see this stems from a vcpkg update. I want to of course make it possible for implot to be used with vcpkg, but supporting pre C++11 compilers is not a burden I'm ready to take on. For future vcpkg updates, I would suggest we ignore or turn off that check if possible. I'm not really familiar with vcpkg so I'll defer to you.

PS -- Thanks for keeping us updated on vcpkg. I'm sure there are several folks who appreciate your efforts!

marcizhu:
macOS developer here 🙋🏻‍♂️

While I have never used vcpkg and thus I don't know exactly how their CI works, I do know for certain that the macOS compiler (Apple clang) defaults to C++98 (just like if you call `g++` without `-std=c++11`) and [the provided log file](https://github.com/microsoft/vcpkg/pull/22988#issuecomment-1032245799) seems to imply they don't explicitly set the C++ version. I guess this could be as easy as adding a `-std=c++11` to the compiler command line on the OSX CI build.

RT222:
Thanks for the explanation. You're right, osx supports new C++ standards, the opposite would have been very surprising actually. vcpkg doesn't care about this anyway, so you are right, the fix was trivial : I just had to specify the desired c++ version in the cmake file.

All is good and all platforms will be supported by the implot vcpkg port in the future, no matter what C++ standard it is targeting. 😃

epezent:
Nice catch, @marcizhu!

Issue #321, opened 2022-02-05
Digital value enhancement


PeterJohnson:
While the digital signal capability of LinePlot is nice, it would be very cool to have a more powerful version of it to display things like digital busses or states.  The input for this would be a X value, text to display, background color, and text color--although some of the examples below would require a boolean for a "empty" state (a middle line).  The output visually would look like some of the examples below.  Ideally it would also support getting hovered state to display extra information.

An enhancement similar to #292 would pair nicely with this, as seen in many of the below examples.

![image](https://user-images.githubusercontent.com/297126/152651936-421140f9-14df-45cf-b831-da8080449d05.png)

![image](https://user-images.githubusercontent.com/297126/152651990-d18f0aa3-f250-4aca-959b-daf6003ffbad.png)

![image](https://user-images.githubusercontent.com/297126/152652062-731b35de-df62-4178-9877-3a73f7ae0b62.png)

![image](https://user-images.githubusercontent.com/297126/152652158-a2cb262d-c6a3-49fd-b3cd-8a6bec5a21f1.png)

epezent:
Hey Peter! Totally agree we should give `PlotDigital` some love. `PlotDigital` has bugged me for a long because of the way it implicitly plots y coordinates in pixel space, and the lack of control we offer over the way it stacks plots vertically. I'd either like to make it behave like the other plots (i.e. y values are in coordinate space), or add some generic functionality to ImPlot that allows the user to set an axis to plot in pixel space (e.g.[ like matplotlib](https://matplotlib.org/2.0.2/users/transforms_tutorial.html)).

Regarding you other points:

> The input for this would be a X value, text to display, background color, and text color--although some of the examples below would require a boolean for a "empty" state (a middle line).

I have generally avoided cluttering `PlotX` functions with additional buffers for per-element colors, text, etc. in an attempt to keep the API concise and maintainable. But obviously that is starting to limit us in more ways that one (e.g. see #298 and my comment there). I am beginning to lean toward adopting an optional buffering API (e.g. `BufferColors(const ImVec4* cols, int count);` that would be called just before `PlotX`. Would love some feedback in this regard!

Text is an odd one, and maybe a callback is best for this. 

> The output visually would look like some of the examples below.

Plot digital could take an `ImPlotDigitalFlags` to specify the styling to be used (FYI, item flags is currently in the works for all plot types in #319)

> Ideally it would also support getting hovered state to display extra information.

[...]

Issue #298, opened 2021-11-18
Scatter plot with different sizes and colors


gorbatschow:
Hello! Just proposal.
It would be great to be able to set sizes and colors as well of scattered points.
![scatexample](https://user-images.githubusercontent.com/33801785/142381107-88c1abf3-373c-4f22-a0c3-79778afeb211.png)

epezent:
Thanks, this is definitely a goal of mine. I haven't quite figured out how to design the API, though. As it is currently designed, the API for styling scatter plots (i.e. `SetNextMarkerStyle`) and providing the data to scatter plots (i.e.. `PlotScatter`) is decoupled. I think this is a good approach for the sake of keeping the arguments passed to `PlotScatter` at a minimum. It gets tricky though if we want to enable per-element styling (this goes for other plot types as well). That would necessarily require `ImU32* colors`, `float* sizes` arguments somewhere. I'm not sure I'm ready to add them to `PlotScatter`, but `SetNextMarkerStyle(ImU32* colors, float* sizes, int count)` feels like a bad idea since it would require either caching the data or requiring the pointers to persist until `PlotScatter`. 

I'm open to any suggestions!

phraktle:
+1 for this feature. From an API user standpoint, it would be convenient to supply additional arrays to `plotScatter` with the size / color overrides.

n00bmind:
+1 I'd like to be able to do something like this too.. An idea that springs to mind is have a new override for plot types that support per-element styling accepting a new type of callback returning a point + styling attributes, i.e. something like:

```
struct ImPlotStyledPoint
{
    double x, y;
    ImU32 col;
    // [...] potentially other attributes, with defaults to indicate 'unused'
};
typedef ImPlotStyledPoint (*ImPlotStyledGetter)(int idx, void* user_data);

IMPLOT_API void PlotScatterG(const char* label_id, ImPlotStyledGetter getter, void* data, int count, ImPlotScatterFlags flags=0);
[...]
```

brenocq:
Happy to announce that this feature was implemented in #672! There's a scatter plot demo with different sizes/color per marker.

<img width="600" alt="image" src="https://github.com/user-attachments/assets/fdd61d1e-9621-4c1f-b9e8-5bd2d7cd2d45" />

PR #332, opened 2022-02-21
add SetNextColorsData to give per point colors

niavok:
Hello,

I needed to have color per point for lines and markers so I developed the feature .

image

This request has already been discuted here :
https://github.com/epezent/implot/issues/298
https://github.com/epezent/implot/issues/321#issuecomment-1037905109

The patch add only one function allowing to override the color used for the data:

`// Set color buffer for the next item only. The colors array must be keep alive until the next item plot.`
`IMPLOT_API void SetNextColorsData(ImPlotCol_ target, ImU32* const& colors, int stride = sizeof(ImU32));`

I know you are not convinced by the need of implicit lifetime management of the color pointer but with 2 calls there is not lot of option without data copy.
As the SetNextColorsData call is often done just before the plot call this liftime management is not so dangerous for most developers. If you find it too dangerous, it could be moved in implot_internal.h

The internal implementation mainly replace some "ImU32 col" by some "const ImU32* col_ptr, int col_stride" parameter. This allow to use a stride of 0 to support easily the legacy behavior with the Style color in the same code path.

I tested and added demo for lines, scatter, VLines and HLines but it should work for other plots.

For the line plot, the color of the first point is used for the the whole line instead of a gradient between the both point color. This is because the AddLine of the ImGui API take only 1 color even if deep inside the draw API, in ImDrawList::AddPolyline, 2 different color could have been provided leading to a gradient in most renderers:

`for (int i = 0; i < points_count; i++)`
`{`
                    `_VtxWritePtr[0].pos = temp_points[i * 2 + 0]; _VtxWritePtr[0].uv = tex_uv0; _VtxWritePtr[0].col = col; // Left-side outer edge`
                    `_VtxWritePtr[1].pos = temp_points[i * 2 + 1]; _VtxWritePtr[1].uv = tex_uv1; _VtxWritePtr[1].col = col; // Right-side outer edge`
                    `_VtxWritePtr += 2;`
`}`

If you have better idea of API or implementation, i can rework the patch, or improve the demo if needed. As the patch modify the internal API of implot, it will be a lot easier to keep sync with the feature directly in the upstream.

Thanks !
Frédéric

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -1117,6 +1117,8 @@ IMPLOT_API void SetNextFillStyle(const ImVec4& col = IMPLOT_AUTO_COL, float alph
 IMPLOT_API void SetNextMarkerStyle(ImPlotMarker marker = IMPLOT_AUTO, float size = IMPLOT_AUTO, const ImVec4& fill = IMPLOT_AUTO_COL, float weight = IMPLOT_AUTO, const ImVec4& outline = IMPLOT_AUTO_COL);
 // Set the error bar style for the next item only.
 IMPLOT_API void SetNextErrorBarStyle(const ImVec4& col = IMPLOT_AUTO_COL, float size = IMPLOT_AUTO, float weight = IMPLOT_AUTO);
+// Set color buffer for the next item only. The colors array must be keep alive until the next item plot.
+IMPLOT_API void SetNextColorsData(ImPlotCol_ target, ImU32* const& colors, int stride = sizeof(ImU32));
 
 // Gets the last item primary color (i.e. its legend icon color)
 IMPLOT_API ImVec4 GetLastItemColor();
--- a/implot_demo.cpp
+++ b/implot_demo.cpp
@@ -960,6 +960,186 @@ void Demo_MarkersAndText() {
     }
 }
 
+void Demo_PlotColors() {
+    // Lines
+    {
+        static float xs1[1001], ys1[1001];
+        ImU32 cs1[1001];
+        for (int i = 0; i < 1001; ++i) {
+            xs1[i] = i * 0.001f;
+            ys1[i] = 0.5f + 0.5f * sinf(50 * (xs1[i] + (float)ImGui::GetTime() / 10));
+            ImVec4 lineColor = ImPlot::SampleColormap(ys1[i], ImPlotColormap_Spectral);
+            lineColor.w = 1 - powf(1 - (i * 0.001f), 4);
+            cs1[i] = ImGui::GetColorU32(lineColor);
+        }
+        static double xs2[11], ys2[11];
+        ImU32 cms2[11];
+        for (int i = 0; i < 11; ++i) {
+            xs2[i] = i * 0.1f;
+            ys2[i] = xs2[i] * xs2[i];
+            cms2[i] = ImGui::GetColorU32(ImPlot::RandomColor());
+        }
+        if (ImPlot::BeginPlot("Line Plot with colors")) {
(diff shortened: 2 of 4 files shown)


epezent:
First, this is amazing work! Thank you for opening this PR

> I know you are not convinced by the need of implicit lifetime management of the color pointer but with 2 calls there is not lot of option without data copy.

We should discuss this in the context of the proposal in #330. It's likely in the near future that we lump styling parameters in the `PlotX` functions and deprecate the decoupled `SetNext` and `Push/Pop` item styling API. 

> For the line plot, the color of the first point is used for the the whole line instead of a gradient between the both point color. This is because the AddLine of the ImGui API take only 1 color even if deep inside the draw API, in ImDrawList::AddPolyline, 2 different color could have been provided leading to a gradient in most renderers:

I plan to eleminate the ImGui fallback implementation in #319  -- it only exists to provide an AA option for line plot (at some point we will add back AA via `RendererLineStripAA`) So, I think it makes more sense to have individual line segments form a gradient between end points.

epezent:
Scanning the code, the first idea that jumps out to me is possibly putting color indexing/getting into the existing `Getter` structs. These currently return an `ImPlotPoint` (in most cases), but perhaps they should return a point-color tuple. Or provide separate `GetData()` and `GetColor()` functions. The point is to keep the indexing/striding implementation out of `RendererLineStrip` etc. 

I'm curious how this should behave for `PlotShaded` and we defintely want to make it workf or `PlotBars` too. 

Also, I think this would also need to honor `offset` as well as `stride`.

niavok:
Thanks for the feedback.

I will wait with attention the new API.

niavok:
Hello,
I updated my code for personal usage and with the new render primitive it was almost a full rewrite.
Now, it used internally a ColorGetter.
The new draw primitive allow one color per vertex so I added color interpolation.
I also added demo for shaded plots and stairstep.
image
image

dnedry2:
I ran into this same issue with my own project. What I did was create additional PlotX functions that take a color array as well as data:
```
void PlotScatter(const char* label_id, const T* xs, const T* ys, ImU32* colormap, bool* mask, int count, int offset=0, int stride=sizeof(T));
void PlotLine(const char* label_id, const T* xs, const T* ys, ImU32* colormap, bool* mask, int count, int offset=0, int stride=sizeof(T));
void PlotShaded(const char* label_id, const T* xs, const T* ys, ImU32* colormap, bool* mask, int count, double y_ref=0, int offset=0, int stride=sizeof(T));
```
This way has the point color as part of ImPlotPoint and the existing getters fetch them. (and a bool mask, to filter out points)


Here's an example of scatter, lines, and shaded lines:
Screen Shot 2022-11-29 at 19 45 34
The two plots share a colormap, but the bottom is far more dense. I interpolate its colormaps myself before calling PlotLines.

Another consideration is the legend colors. I'm not sure how to decide that with per point color.

mihaly-sisak:
Is there a possibility of this being merged? I just want to plot color data from circular buffers :C
What can I do to help the color functions to be included in ImPlot?

brenocq:
Hi @niavok @mihaly-sisak @dnedry2! Since `ImPlotSpec` (#519) is now merged into main, I reimplement this per point colors in a new PR #672 by following the `ImPlotSpec` style. I would love some feedback on #672!

brenocq:
I'm closing this PR since this feature was implemented in #672. @niavok your implementation was an inspiration for #672, thank you!!

PR #389, opened 2022-08-18
Demo of Largest Triangle Three Buckets (LTTB) Downsampling Algorithm

ozlb:
"Downsampling time series for visual representation" by Sveinn Steinarsson.
https://skemman.is/bitstream/1946/15343/3/SS_MSthesis.pdf
https://github.com/sveinn-steinarsson/flot-downsample

### DIFF
--- a/implot_demo.cpp
+++ b/implot_demo.cpp
@@ -90,23 +90,23 @@ ImVec4 RandomColor() {
 }
 
 double RandomGauss() {
-	static double V1, V2, S;
-	static int phase = 0;
-	double X;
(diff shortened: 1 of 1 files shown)


epezent:
@ozlb -- I agree this should be demoed. I would prefer to see a separate demo that is more visible though (e.g. under `Tools/Downsampling` or `Custom/Downsampling` or a new tab altogether). It would also be nice to make a generic LTTB downsampler class that folks can copy for their own applications (e.g. something like `ScrollingBuffer` which is frequently copy/pasted by our users).

PR #393, opened 2022-08-24
Continuous integration (CI) workflow: build the code in your repository and run tests

ozlb:
@epezent this is an idea, probably cmake is more modern and clean solution.

CI using GitHub Actions offers workflows that can build the code in your repository and run your tests.
GitHub runs your CI tests and provides the results of each test in the pull request, so you can see whether the change in your branch introduces an error. When all CI tests in a workflow pass, the changes you pushed are ready to be reviewed by a team member or merged. When a test fails, one of your changes may have caused the failure.
More about "runners"... https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

i.e. workflow run report in my fork
build_implot

PR #395, opened 2022-09-01
CI setup

rokups:
I would like to propose a more complete CI setup than #393. I used a minimal CMake script to simplify building across multiple platforms. This script is treated purely as CI facility script and is put to `.github` subdir in order to discourage users from using it.

Note that i included my PR #392 in this PR. I enabled max warning level on all platforms, so in order for build to succeed i needed those warnings fixed.

Tested platforms: Windows / Linux / MacOS
Tested compilers: msvc / mingw / gcc / clang
Tested architectures: x86 / x64
Sample build: https://github.com/rokups/implot/actions/runs/2972494310

### DIFF
--- a/.github/CMakeLists.txt
+++ b/.github/CMakeLists.txt
@@ -1,0 +1,25 @@
+# This build script is not meant for general use, it is for CI use only!
+cmake_minimum_required(VERSION 3.0)
+project(implot)
+
+# Same as Dear ImGui
+set(CMAKE_CXX_STANDARD 11)
+
+file(GLOB SOURCE_CODE ../implot*.*)
+add_library(implot STATIC ${SOURCE_CODE})
+
+if(MSVC)
+    target_compile_options(implot PRIVATE /W4 /WX)
+else()
+    target_compile_options(implot PRIVATE -Wall -Werror -pedantic)
+endif()
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU" AND DEFINED GCC_ARCH)
+    if ("${GCC_ARCH}" MATCHES "Win32|x86|32")
+        target_compile_options(implot PRIVATE -m32)
+    elseif ("${GCC_ARCH}" MATCHES "Win64|x64|64")
+        target_compile_options(implot PRIVATE -m64)
+    endif ()
+endif ()
+
+target_include_directories(implot PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../imgui)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,113 @@
+name: build
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  Linux:
+    runs-on: ubuntu-22.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        build_type:
+          - debug
+          - release
+        compiler:
+          - gcc
+          - clang
+        arch:
+          - x86
+          - x64
+
+    steps:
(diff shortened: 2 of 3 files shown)


epezent:
Thanks Rokas! This looks great. This is definetely something I've wanted to do but have not had time. So I'm more than happy to accept your solution. Since I am more familiar with CMake, I feel better merging this PR than #393 in case I need to make updates later down the road (@ozlb, thank you for your submission too!)

I don't have any immediate questions or concerns, so I will go ahead and merge this work.

epezent:
Actually, I have one question -- is it possible to setup the CI so that jobs are triggered when changes are made to ocornut/imgui (so that we are covered by our changes and ImGui's changes)?

rokups:
Once test engine is public i will also add some muscle-flexing to the CI, basic tests that expand every demo example :rocket: 

Thing with testing upstream commits is a little complicated. We have two options:
1. Have ocornut/imgui repo send a workflow dispatch event to epezent/implot and do builds on every push to upstream.
2. Do CI builds on schedule, say every day or every week for example.

1st option sounds perfect, but recently i could not get it working, even though i did get it working in the past. Either i missed something obscure, or github changed their API and broke it. I will be looking into it once again later and if i succeed i will contribute it to implot.

2nd option is very simple to implement. All you have to do is to add `on: schedule` section in yml:
```yml
on:
  <other stuff here>
  schedule:
    - cron: '40 8 * * *'
```
It follows standard cron syntax. Downside is of course that builds will run even when there are no commits pushed to upstream master branch for multiple days.

ocornut:
Thanks for merging !

I personally wasn't too sure about the cmake bits and suggested to rokups to try a version without, which you can see here: 
https://github.com/epezent/implot/commit/c76c88f6727d6457c9c05c70adf67adea85923a2
I personally think this is slightly better but Rokas disagree. I guess they both are nice, just wanted to point out this commit existed somewhere.

One thing I would suggest adding is an extra build with `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` and `IMGUI_DISABLE_OBSOLETE_KEYIO` defined. I don't think it needs to be done for all variants and compilers, this is merely a way for ImPlot to catch on those changes.

ImPlot being in the particular case of being a non-trivial yet commonly used extension, I guess it should keep putting some effort at using IMGUI_VERSION_NUM checks here and there, but you might eventually also consider to drop some code-path after a given time (e.g. I'm using 2 years).

KarKarKarKarkarkar:
@ocornut 


> Thanks for merging !
> 
> I personally wasn't too sure about the cmake bits and suggested to rokups to try a version without, which you can see here: [c76c88f](https://github.com/epezent/implot/commit/c76c88f6727d6457c9c05c70adf67adea85923a2) I personally think this is slightly better but Rokas disagree. I guess they both are nice, just wanted to point out this commit existed somewhere.
> 
> One thing I would suggest adding is an extra build with `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` and `IMGUI_DISABLE_OBSOLETE_KEYIO` defined. I don't think it needs to be done for all variants and compilers, this is merely a way for ImPlot to catch on those changes.
> 
> ImPlot being in the particular case of being a non-trivial yet commonly used extension, I guess it should keep putting some effort at using IMGUI_VERSION_NUM checks here and there, but you might eventually also consider to drop some code-path after a given time (e.g. I'm using 2 years).

https://github.com/ocornut/imgui/issues/7117

Autofocus random window no in Windows OS, and other graphic interfaces. You need to.remove this overcoding or adding option to disable this useless feature.

Issue #379, opened 2022-07-07
Support for long double ?


pthom:
Hello,

I found that the various IMPLOT_TMP template implementations do not support `long double`.

See extract from `implot_items.cpp`:

````cpp
template <typename T>
void PlotLine(const char* label_id, const T* xs, const T* ys, int count, ImPlotLineFlags flags, int offset, int stride) {
    GetterXY<IndexerIdx<T>,IndexerIdx<T>> getter(IndexerIdx<T>(xs,count,offset,stride),IndexerIdx<T>(ys,count,offset,stride),count);
    PlotLineEx(label_id, getter, flags);
}

template IMPLOT_API void PlotLine<ImS8>(const char* label_id, const ImS8* xs, const ImS8* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImU8>(const char* label_id, const ImU8* xs, const ImU8* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImS16>(const char* label_id, const ImS16* xs, const ImS16* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImU16>(const char* label_id, const ImU16* xs, const ImU16* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImS32>(const char* label_id, const ImS32* xs, const ImS32* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImU32>(const char* label_id, const ImU32* xs, const ImU32* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImS64>(const char* label_id, const ImS64* xs, const ImS64* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<ImU64>(const char* label_id, const ImU64* xs, const ImU64* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<float>(const char* label_id, const float* xs, const float* ys, int count, ImPlotLineFlags flags, int offset, int stride);
template IMPLOT_API void PlotLine<double>(const char* label_id, const double* xs, const double* ys, int count, ImPlotLineFlags flags, int offset, int stride);
````

"long double" is rather unusual, but it might help some people. 

[...]

PR #397, opened 2022-09-07
Support long & long double, add macro INSTANTIATE_FOR_NUMERIC_TYPES (Fix #319)

pthom:
This fixes the issue [#319](https://github.com/epezent/implot/issues/379) which stated that long double are not supported. 

As a matter of fact, `long double` as well as `unsigned long` and `signed long` are not supported when using dynamic linking: see demo repository for a minimal reproduction here : https://github.com/pthom/tmp_implot_dyn

#### Commit details

There are two commits: 

- [the first commit](https://github.com/pthom/implot/commit/df1a5a2f7664226e32a53c4f0da8cbcf7a0d96ca) introduces a macro `INSTANTIATE_FOR_NUMERIC_TYPES` that reduces some boilerplate code around template instantiations. It provides a "setting" (`IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES`) in order to enable / disable the addition of "long double" & "signed/unsigned long". 
[All the CI tests](https://github.com/pthom/implot/actions/runs/3007725925) succeed.
The compilation time of implot_items.cpp went up from 3.2 seconds to 4 seconds

- [the second commit](https://github.com/pthom/implot/commit/47409a46d13df801cb1072454f154142364fb3be) disables this setting (and it can be activated by the user via the compil time define `IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES`)
[All the CI tests](https://github.com/pthom/implot/actions/runs/3007735577) succeed.

####  Notes about numeric types ~~synonyms~~ quasi-synonyms:

  On some platforms, "unsigned long" might be the same size as "unsigned long long", but it is nonetheless a separate type: see https://godbolt.org/z/1KWv5re7q (example with GCC 64 bits)

  On some other platforms, "long double" might be the same size as "double", but it is nonetheless a separate type: see https://godbolt.org/z/ae71P7rqG (example with MSVC 64 bits)

---

#### Context: 

I discovered this issue while creating python bindings for imgui and implot. Since numpy uses "long double" as the default float type, and "long" as the default integer type, python code like this would fail under linux 64 bits:

````python
xs = np.array((1, 2, 3, 4))
implot.PlotBars("Bars", xs)  # xs is of type "signed long"
```` 
 
For info, the bindings I'm developing are not ready for production (hopefuly they will soon).  However my intent is to write an automatic binding generation tool that takes great care in reproducing the layout and the documentation found in the library header files. See [implot.pyi](https://github.com/pthom/lg_imgui_bundle/blob/main/bindings/lg_imgui_bundle/implot.pyi#L1194) and [imgui.pyi](https://github.com/pthom/lg_imgui_bundle/blob/main/bindings/lg_imgui_bundle/imgui.pyi). For the gory details, the generated code automatically transcribes between a numpy ndarray and the corresponding C++ buffer, such as here for the [plot_line function](https://github.com/pthom/lg_imgui_bundle/blob/main/bindings/pybind_implot.cpp#L589)

### DIFF
--- a/.github/CMakeLists.txt
+++ b/.github/CMakeLists.txt
@@ -23,3 +23,4 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU" AND DEFINED GCC_ARCH)
 endif ()
 
 target_include_directories(implot PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../imgui)
+target_compile_definitions(implot PRIVATE IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES=1)
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ A: Yes, within reason. You can plot tens to hundreds of thousands of points with
 **Q: What data types can I plot?**
 
 A: ImPlot plotting functions accept most scalar types:
-`float`, `double`, `int8`, `uint8`, `int16`, `uint16`, `int32`, `uint32`, `int64`, `uint64`. Arrays of custom structs or classes (e.g. `Vector2f` or similar) are easily passed to ImPlot functions using the built in striding features (see `implot.h` for documentation), and many plotters provide a "getter" overload which accepts data generating callbacks.
+`float`, `double`, `int8`, `uint8`, `int16`, `uint16`, `int32`, `uint32`, `int64`, `uint64`. Arrays of custom structs or classes (e.g. `Vector2f` or similar) are easily passed to ImPlot functions using the built-in striding features (see `implot.h` for documentation), and many plotters provide a "getter" overload which accepts data generating callbacks. Additional support for `long`, `unsigned long` and `long double` can be added by defining `IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES` at compile-time (see `implot_items.cpp`).
 
 **Q: Can plot styles be modified?**
 
--- a/implot_internal.h
+++ b/implot_internal.h
@@ -167,7 +167,7 @@ static inline double ImMean(const T* values, int count) {
     double den = 1.0 / count;
     double mu  = 0;
     for (int i = 0; i < count; ++i)
-        mu += values[i] * den;
+        mu += (double)values[i] * den;
     return mu;
 }
 // Finds the sample standard deviation of an array
@@ -177,7 +177,7 @@ static inline double ImStdDev(const T* values, int count) {
     double mu  = ImMean(values, count);
     double x   = 0;
     for (int i = 0; i < count; ++i)
-        x += (values[i] - mu) * (values[i] - mu) * den;
+        x += ((double)values[i] - mu) * ((double)values[i] - mu) * den;
     return sqrt(x);
 }
 // Mix color a and b by factor s in [0 256]
--- a/implot_items.cpp
+++ b/implot_items.cpp
@@ -93,7 +93,7 @@ static IMPLOT_INLINE float  ImInvSqrt(float x) { return 1.0f / sqrtf(x); }
 // Uncomment the next line (#define IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES) in order to define versions for those types
 // (or define it via a compilation option). In this case, the compilation time for this specific file will be 33% longer
 
-#define IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES
+// #define IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES
 
 #define INSTANTIATE_FOR_REMAINING_NUMERIC_TYPES(instantiate_macro) \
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

**Comment by pthom** [comment]
I created a minimal repository that show a link error when using `ImPlot::PlotLine<unsigned long>` in the case of a shared library, here: https://github.com/pthom/tmp_implot_dyn

**Comment by epezent** [comment]
Today I learned you can pass a macro function to a macro function. This is awsome! I have wanted to do exactly what  `INSTANTIATE_FOR_NUMERIC_TYPES` but never figured out how. 

I have a few suggestions:

1. change `instantiate_PlotLine` to `INSTANTIATE_PLOT_LINE` (personal preference)
2. collapse `instantiate_PlotLine`, `instantiate_PlotLine2`, etc. to one macro (feels cleaner to not have numbered macros)

```cpp
#define INSTANTIATE_PLOT_LINE(T) \
template IMPLOT_API void PlotLine<T> (const char* label_id, const T* values, int count, double xscale, double x0, ImPlotLineFlags flags, int offset, int stride); \
template IMPLOT_API void PlotLine<T>(const char* label_id, const T* xs, const T* ys, int count, ImPlotLineFlags flags, int offset, int stride);

INSTANTIATE_FOR_NUMERIC_TYPES( INSTANTIATE_PLOT_LINE );
```

pthom:
Thanks for your review! You are right, the preprocessor is almost a second awkward hidden language inside C++, after the templates ;-) 

I pushed your requested changes, they do improve the code.

There is one more thing I'd like to discuss. As a matter of fact, it also fails with static builds (see https://github.com/pthom/tmp_implot_dyn and edit the CMakeList to set `set(BUILD_SHARED_LIBS OFF)`. 

Thus, the setting that I added might seem superfluous,  since I think "signed long" and "unsigned long" are quite common and should be supported. What do you think?

epezent:
I think it's ok to require the user to define `IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES` for these types (for both static and dynamic linkage). Even though they are common, no one has every requested support for them until now. I'd rather not add these as default types and further extend compile times. 

I'd even argue we may want to move some already existing types behind `IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES` to further reduce compile times. In #328 I suggested only instantiating `int`, `float`, and `double` by default. But we can make this decision at a later time. 

I would suggest adding the following to the end of `.github/CMakeLists.txt`:

`target_compile_definitions(implot PRIVATE IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES=1)`

pthom:
OK, I added the definition in the CMakeList.txt.

In order to prepare for a merge, I also reworded the comments on one line in implot_items.cpp.

Also, I mentioned of this option in the FAQ part of the Readme (**Q: What data types can I plot?**)

ocornut:
Can this setup be used to manually restrict instanciations to seleected types? Eg embedded apps could instanciate only 1-2 types.

epezent:
> Can this setup be used to manually restrict instanciations to seleected types? Eg embedded apps could instanciate only 1-2 types.

I'd be interested in supporting this. Probably we'd want somehing like `IMPLOT_INSTANTIATE_FOR_FLOAT`, `IMPLOT_INSTANTIATE_FOR_DOUBLE`, etc. which default to the current behavior unless explicitly defined by the user. @pthom , do you see any way to extend this PR to support something like that? I'm not seeing an immediate solution that wouldn't require multiple preprocessor passes (i.e. we can't just put `#if IMPLOT_INSTANTIATE_FOR_FLOAT` etc. in the body of `INSTANTIATE_FOR_STANDARD_NUMERIC_TYPES`)

ocornut:
I recall there’s some sort of trick you can use to provide a list of items (here types) in a macro and then expand this list for other macros.

pthom:
@ocornut @epezent : would something like [this](https://github.com/pthom/implot/commit/f059897668ebd7502d73aaa57e424f4ce90f83f2) be OK?

It can be summarized as:

__implot_config.h__

````cpp
 #define IMPLOT_CUSTOM_NUMERIC_TYPES(m) m(int); m(float); m(double);
````

__implot_items.cpp__
````cpp
ifdef IMPLOT_CUSTOM_NUMERIC_TYPES
#define INSTANTIATE_FOR_NUMERIC_TYPES IMPLOT_CUSTOM_NUMERIC_TYPES
#else
... same as before
````

This solution forces the addtions of a `implot_config.h` file, since function-like macros cannot reliably be passed via the command line (and they are incompatible with cmake).

However, it is quite readable, and quite easy to customize. 

I think we should try to stay simple with the preprocessor, because I suspect that more work on list expansion would require bad looking tricks that resemble [this](https://stackoverflow.com/questions/49235603/c-preprocessor-iteratively-expand-macro-to-comma-separated-list).

PR #402, opened 2022-09-14
Remove support for IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES

pthom:
> My two cents: given full configurability (looks great!) i would suggest not adding “ IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES. For those rare types it seems reasonable to simply let the user express their needs explicitly.”

I do agree, and this PR removes them. 

--- 

> especially as it may be tempting for users and we don’t know what long double would carry in term of math runtimes. 

That is an interesting question! I made a quick study, and my findings are:

- the difference of speed between `float` and `double` can be as high as 20% (and the result do depend on the arch and compiler)
- when `double` and `long double` have the same size (8 bytes), `long double` is 20% to 30% slower (except on apple M1 processor)
- when `sizeof(long double) = 16` vs `sizeof(double) = 8`, long double can be 10 to 30 times slower (this depends on the compiler and platform)
--- 

> I haven’t poked in internals but wondered if small types <32 or <64 could somehow attempt to reuse larger types functions, with special adapters? Intuitively it seems like u8 u16 could use the code for u64. This is implying that generated code size and link size could be reduced. With custom type list this is now perhaps a little harder to setup, so perhaps it would need an hardcoded proof of concept because getting it to work with custom type lists.

There is perhaps a solution, but I suspect it would be very likely be "platform / compiler / arch / type" specific and maybe difficult to maintain. 

In the case of unsigned integers, a possible solution would be to:
- implement only the full code for the u64 version
- have a template function for u8, u16, u32 that would call the u64 version by first reading elements as u64.
- have a way to read u64 values inside a u8, u16 or u32 array. A possible solution would be to set  `u64 value = 0;`, then fill the right portion of `value`'s bytes with the bytes from the array

However, based on my observations, the memory layout for a 64 bits int (on a Mac M1) is a mix of:
- little-endian (for bits)
- big-endian (between bytes)
- big-endian (between groups of 4 bytes)
And I suspect that this is very much platform dependent.

Example, if I set 
`int64_t v64 = 0x123456789ABCDE00;` 
and examine the memory at &v64, it looks like this on my Mac:
````
    00 de bc 9a   78 56 34 12
````

### DIFF
--- a/.github/CMakeLists.txt
+++ b/.github/CMakeLists.txt
@@ -71,22 +71,21 @@ if (UNIX)
     target_link_libraries(implot PUBLIC m stdc++)
 endif()
 
-# Switch between several sets of numeric types (by adding `-DIMPLOT_NUMERIC_SET=default|custom|all`)
-if (DEFINED IMPLOT_NUMERIC_SET)
-    if ("${IMPLOT_NUMERIC_SET}" STREQUAL "default")
-        message(STATUS "Compiling for default types")
-    elseif("${IMPLOT_NUMERIC_SET}" STREQUAL "custom")
-        message(STATUS "Compiling for custom types (float and double)")
-        target_compile_definitions(implot PRIVATE "IMPLOT_CUSTOM_NUMERIC_TYPES=(float)(double)")
-    elseif("${IMPLOT_NUMERIC_SET}" STREQUAL "all")
-        message(STATUS "Compiling for all types")
-        target_compile_definitions(implot PRIVATE "IMPLOT_INSTANTIATE_ALL_NUMERIC_TYPES=1")
-    else()
-        message(FATAL_ERROR "unhandled IMPLOT_NUMERIC_SET=${IMPLOT_NUMERIC_SET}")
-    endif()
+# Define supported types via command line:
+#   - With no choice all types are supported (so that the CI provides support for all known types)
+#   - with -DIMPLOT_CUSTOM_NUMERIC_TYPES="default" the default set defined in implot_items.cpp is used
+#   - with -DIMPLOT_CUSTOM_NUMERIC_TYPES="(int)(float)(double)" only int, float and double are supported
+if (NOT DEFINED IMPLOT_CUSTOM_NUMERIC_TYPES)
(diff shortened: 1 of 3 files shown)


pthom:
If you are interested, here are some extracts from my study on floating types performances:


__With [x64 msvc v19.latest](https://godbolt.org/z/43jh8P4EG)__
> double is faster than float on windows 64 bits builds. long double, although having the sizeof(double), is a little slower than double.

    float size=4
    double size=8
    long double size=8
    Function            |Nb calls|Total time|Av. time |Deviation|
    --------------------+--------+----------+---------+---------+
    exercise_long double|     100|  19.754ms|197.539us|  8.507us|
    exercise_float      |     100|  16.402ms|164.025us|234.326us|
    exercise_double     |     100|  14.118ms|141.179us|  8.176us|

__With [x86-64 clang 15.0.0](https://godbolt.org/z/r9ed48GfY)__
> double is 20% slower than float. long double (16 bytes) is _18 times slower_ than double

    float size=4
    double size=8
    long double size=16
    Function            |Nb calls|Total time|Av. time  |Deviation|
    --------------------+--------+----------+----------+---------+
    exercise_long double|     100| 217.883ms|2178.832us|112.920us|
    exercise_double     |     100|  12.327ms| 123.269us|  4.157us|
    exercise_float      |     100|  10.486ms| 104.858us| 85.306us|

__With [x86-64 gcc 12.2](https://godbolt.org/z/1PrqGaaf4)__
> double is 1.8 times slower than float! long double (16 bytes) is _10 times slower_ than double.

    run_benchmark on float : size=4
    run_benchmark on double : size=8
    run_benchmark on long double : size=16
    Function            |Nb calls|Total time|Av. time  |Deviation|
    --------------------+--------+----------+----------+---------+
    exercise_long double|     100| 211.380ms|2113.799us|148.406us|
    exercise_double     |     100|  19.671ms| 196.711us|  8.312us|
    exercise_float      |     100|  11.958ms| 119.579us| 17.470us|

PR #407, opened 2022-09-26
Fixes for ImGuiModFlags_XXX -> ImGuiMod_XXX (v1.89) (#347)

ocornut:
This is a followup to #347 and an update following this change in core lib:
https://github.com/ocornut/imgui/commit/fd408c97904fbf9f3b4c786318e8389de11bc681

Like with #347 this is only required for users building with `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` + aiming to support a few older versions.

----

If you don't mind raising minimum imgui version to latest, you can start taking advantage of the fact that ImGuiMod_XXX values can be safely or-ed with an ImGuiKey, so stuff like:

```
ImGuiMouseButton Pan;           // LMB    enables panning when held,
ImGuiModFlags    PanMod;        // none   optional modifier that must be held for panning/fitting
```

Could become:
```
ImGuiKeyChord Pan;    // == ImGuiKey_MouseLeft | ImGuiMod_Ctrl
```
(adding Ctrl for the example but PanMod default to None).

I would however suggest not making those changes yet, wait until at least 1.89 releases.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -4635,32 +4635,32 @@ ImPlotInputMap& GetInputMap() {
 void MapInputDefault(ImPlotInputMap* dst) {
     ImPlotInputMap& map = dst ? *dst : GetInputMap();
     map.Pan             = ImGuiMouseButton_Left;
-    map.PanMod          = ImGuiModFlags_None;
+    map.PanMod          = ImGuiMod_None;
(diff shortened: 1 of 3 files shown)


epezent:
Nice! I'll have to look into this and see if it's worth raising the min version (my hunch is not right now, but maybe in the future). Thanks for the PR!

PR #412, opened 2022-10-11
Small Pie Chart enhancements

Ben1138:
- Added `ImPlotPieChartFlags_IgnoreHidden` flag
- Added `PlotPieChart` overload that accepts an `ImPlotFormatter`.

I especially needed the `IgnoreHidden` feature for my use case, so I figured why not share them. Let me know what you think.

piechartenhance

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -287,8 +287,9 @@ enum ImPlotInfLinesFlags_ {
 
 // Flags for PlotPieChart
 enum ImPlotPieChartFlags_ {
-    ImPlotPieChartFlags_None      = 0,      // default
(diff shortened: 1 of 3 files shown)


epezent:
Hey! Apologies for the slow response. Love the PR. My only recommendation is that we make the `const char* fmt` overload call your new `ImPlotFormatter` overload under the hood to reduce repeated code. Do you mind making that change?

Ben1138:
Done

epezent:
@Ben1138 , sincere apologies for not merging this sooner. Would still like to, if you'd be willing to fix the conflicts.

Ben1138:
Done

Issue #421, opened 2022-11-11
sprintf is unsafe


mcmara:
Despite #310, there are still instances of sprintf in ```v0.14```. Xcode 14.1 with ```-Wall -Werror``` fails to build it with tons of errors:
```
[... cut ...]
[ 96%] Building CXX object lib/CMakeFiles/implot-opengl.dir/__/implot/implot_demo.cpp.o
[ 96%] Building CXX object lib/CMakeFiles/implot-opengl.dir/__/implot/implot_items.cpp.o
[ 96%] Building CXX object lib/CMakeFiles/implot-opengl.dir/__/implot/implot.cpp.o
/path/to/implot/implot_demo.cpp:840:17: error: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Werror,-Wdeprecated-declarations]
                sprintf(label, "digital_%d", i);
                ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/sys/cdefs.h:215:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
                                                      ^
/path/to/implot/implot_demo.cpp:846:17: error: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Werror,-Wdeprecated-declarations]
                sprintf(label, "analog_%d", i);
                ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
[...]

PR #426, opened 2022-11-24
Use snprintf instead of sprintf in implot_demo.cpp

jminor:
sprintf is marked as deprecated and/or unsafe by some compilers. For example, clang complains:
```
implot_demo.cpp:840:17: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
```

This PR replaces all calls to `sprintf` in `implot_demo.cpp` with `snprintf` to address this concern.

Addresses #421

### DIFF
--- a/implot_demo.cpp
+++ b/implot_demo.cpp
@@ -837,13 +837,13 @@ void Demo_DigitalPlots() {
         ImPlot::SetupAxisLimits(ImAxis_Y1, -1, 1);
         for (int i = 0; i < 2; ++i) {
             if (showDigital[i] && dataDigital[i].Data.size() > 0) {
-                sprintf(label, "digital_%d", i);
+                snprintf(label, sizeof(label), "digital_%d", i);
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM. And the length for `label` is also sufficient (would otherwise potentially throw compile warnings depending on `-Wformat=???` setting).

implot_demo.cpp
```diff
@@ -1880,7 +1880,7 @@ void Demo_OffsetAndStride() {
         ImPlot::PushColormap(ImPlotColormap_Jet);
         char buff[16];
```
BenBE, line 1881: Should be at least 20 bytes (8 bytes format text + 10 bytes for `%d`). Could be even 32; to align with buffer size used above.
```suggestion
        char buff[32];
```

PR #458, opened 2023-03-20
Prevent negative resolution being passed to Locator

### LINKED ISSUE #457: Resolution passed to ImPlotLocator can be negative
This is causing my application to hang after minimizing. I think it was because the canvas rect was zero, so after subtracting the padding the `plot_height` would be negative.

torkleyy:
Fixes #457

Not sure if this is the best way to fix the bug, but it solved the freezing for my application.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -2540,7 +2540,7 @@ void SetupFinish() {
     // (3) calc left/right pad
     PadAndDatumAxesY(plot,pad_left,pad_right,gp.CurrentAlignmentV);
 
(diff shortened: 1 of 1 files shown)


epezent:
1) curious what circumstances got you to a zero width plot
2) probably need to do the same for y-axis locators in the block just above this one

epezent:
https://github.com/epezent/implot/commit/b5b796a1d4031c445f61b4fb740d5171d8aa56fd

PR #465, opened 2023-04-09
Embrace 'nullptr'

Minty-Meeo:
Hello.  This PR is the next step of a multi-codebase crusade against using zero as a null pointer constant.  I have made a [PR for Imgui](https://github.com/ocornut/imgui/pull/6313) with similar changes.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -147,7 +147,7 @@ You can read releases logs https://github.com/epezent/implot/releases for more d
 
 // Global plot context
 #ifndef GImPlot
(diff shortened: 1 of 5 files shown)


ocornut:
Thank you!

@epezent I took the liberty to merge this. I contemplated adding the warning pragma/push warning disable as the big files in imgui codebase, but my main reason for not applying the same in imgui codebase don't apply nearly as much here so I thought it would be sane to simply go ahead and use nullptr.

epezent:
Thanks Omar!

PR #481, opened 2023-06-14
drag with entire dragrect

junho7:
Drag with entire dragrect rather than the point in the center


junho7:
I just added a few lines. But diff shows entire file has changed. Let me investigate it

PR #490, opened 2023-07-10
subplotsSizing demo crash correction

GGbaba:


epezent:
fixed in https://github.com/epezent/implot/commit/bfc1e0908e41018c98ad91be47fec846e9d4aea4

PR #507, opened 2023-08-21
Ability to query point, line and rect drags for their click, hovered and held states

pozemka:
Hello again! Following #506 PR.

To be able to query states of draggable items I've added new optional arguments:
```cpp
IMPLOT_API bool DragPoint(int id, double* x, double* y, const ImVec4& col, float size = 4, ImPlotDragToolFlags flags = 0, 
    bool* clicked = nullptr, bool* hovered = nullptr, bool* held = nullptr);
```
When `clicked`, `hovered` or `held` provided it will be set to results of `ImGui::ButtonBehavior` function.

Example code:
```cpp
bool is_clicked = false;
ImPlot::DragPoint(key.id, &x, &y, ImVec4(1.0f, 0.0f, 1.0f, 1.0f), line_weight, 0, &is_clicked);
if (is_clicked)
{
  //process click
}
```

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -4170,9 +4170,11 @@ bool DragRect(int n_id, double* x_min, double* y_min, double* x_max, double* y_m
 
     const bool mouse_inside = rect_grab.Contains(ImGui::GetMousePos());
     const bool mouse_clicked = ImGui::IsMouseClicked(0);
+    const bool mouse_down = ImGui::IsMouseDown(0);
     if (input && mouse_inside) {
         if (out_clicked) *out_clicked = *out_clicked || mouse_clicked;
-        if (out_hovered) *out_hovered = true;        
+        if (out_hovered) *out_hovered = true;
+        if (out_held)    *out_held    = *out_held    || mouse_down;
     }
 
     PushPlotClipRect();
--- a/implot.cpp
+++ b/implot.cpp
@@ -3894,7 +3894,7 @@ IMPLOT_API void TagYV(double y, const ImVec4& color, const char* fmt, va_list ar
 
 static const float DRAG_GRAB_HALF_SIZE = 4.0f;
 
-bool DragPoint(int n_id, double* x, double* y, const ImVec4& col, float radius, ImPlotDragToolFlags flags, bool* is_clicked, bool* is_hovered, bool* is_held) {
+bool DragPoint(int n_id, double* x, double* y, const ImVec4& col, float radius, ImPlotDragToolFlags flags, bool* out_clicked, bool* out_hovered, bool* out_held) {
     ImGui::PushID("#IMPLOT_DRAG_POINT");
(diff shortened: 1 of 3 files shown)


epezent:
Two comments:

1. Let's name the arguments `out_clicked`, `out_hovered`, `out_held` to be consistent with `ImGui::ButtonBehavior`
2. `DragRect` calls `ImGui::ButtonBehavior` button behavior five times -- once for the center, and once for each corner. We need to handle this somehow. A couple options:
    -  change the output args to bool out_xxx[5]
    - aggregate the outputs, e.g. `*is_clicked = *is_clicked || clicked`
    - should report the state change of the center

The first option feels heavy, the last one feels misleading. So perhaps the second option is the best generalization. Also, `DragRect` might want to report "hovered" when the cursor is inside the rect (but not necessarily hovering one of the control points). 

Thoughts?

pozemka:
1. Comments added
2. Arguments renamed
3. `DragRect` is tricky. 🤔 In my opinion most useful would be to know if it was clicked anywhere. Actually I can imagine scenarios when user would be interested in clicks on corner points or edges but then we could also imagine need for a RMB, Scroll etc. So I think it is better to provide facility to handle most common case. 

I've added handling of events on entire rectangle:
```cpp
if (input) {
    // Entire rectangle
    ImGuiID r_id = id + 5;
    ImGui::KeepAliveID(r_id);
    bool is_hovered = false;
    bool is_held = false;
    bool is_clicked = ImGui::ButtonBehavior(rect, r_id, &is_hovered, &is_held);
    if (out_clicked)
        *out_clicked = is_clicked;
    if (out_hovered)
        *out_hovered = is_hovered;
    if (out_held)
        *out_held = is_held;
}
```
But I not sure about `r_id` value for this ButtonBehavior. In the loop earlier it had values `id+1 .. id+4` so `id + 5` seems correct?

Also I have update example code to print info about state of draggable items (only one item per plot though):
image

epezent:
@pozemka, I just pushed a few changes:

1. We can't use a ButtonBehavior for the entire rect because it prevents the mouse from passing through and interacting with the plot. This is problematic if the rect covers or exceeds the plot extents (basically makes the entire plot unusable). But, I agree with you on the rect needing to provide click/hover events for its interior, so I added some custom logic to handle this case. 
2. I updated the demo examples to show something a bit more interesting/practical (changes line thickness when hovered etc.)
3. Just a few preferential changes to the documentation and code. 

Let me know what you think and we'll merge this...

pozemka:
I've looked through your commit.
Everything seems fine. My initial thought was thath this code
```cpp
const bool mouse_inside = rect_grab.Contains(ImGui::GetMousePos());
const bool mouse_clicked = ImGui::IsMouseClicked(0);
if (input && mouse_inside) {
    if (out_clicked) *out_clicked = *out_clicked || mouse_clicked;
    if (out_hovered) *out_hovered = true;        
}
```
Makes previous hover/click checks unnecessary. But parts of the grab points are outside of the rect so they should be handled too.
Only thing I could imagine as missing is that the rect itself has no held state (grab points have it). So I have added `ImGui::IsMouseDown`.

PR #511, opened 2023-08-28
Fix indexing logic of GetterHeatmapColMaj::operator()

cwahn:
Change indexing mod from Cols to Rows to fix bounding point implementation error.


epezent:
Thanks! Clearly a copy-paste oversight.

PR #517, opened 2023-09-25
Fix EndPlot() not deducing auto color for ImPlotCol_PlotBorder

elvissteinjr:
EndPlot() currently accesses the style color field directly when deciding whether to render the border and behaves wrongly if ImPlotCol_PlotBorder is set to IMPLOT_AUTO_COL, never rendering it. Despite being documented to defaulting to ImGuiCol_Border's value.

Of course the easy fix on the application end is to set ImPlotCol_PlotBorder explicitly. This change may also be considered as a breaking change as it will start adding borders to plots that previously had none with default styling.
Changing the documentation to reflect the current behavior could be an alternative if preferred.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -2778,7 +2778,7 @@ void EndPlot() {
 
     // FINAL RENDER -----------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


epezent:
Thanks for the fix!

PR #519, opened 2023-09-30
Add ImPlotSpec and remove existing plot item styling mechanisms

epezent:


epezent:
@PeterJohnson, @sergeyn, @sonoro1234, @rokups, @marcizhu , @ocornut, @bear24rw, @ozlb -- I would appreciate some eyes on this. Please feel free to leave comments or critiques.

implot_internal.h
```diff
@@ -1173,30 +1192,17 @@ struct ImPlotNextPlotData
...
-    float           LineWeight;
-    ImPlotMarker    Marker;
-    float           MarkerSize;
-    float           MarkerWeight;
-    float           FillAlpha;
-    float           ErrorBarSize;
-    float           ErrorBarWeight;
-    float           DigitalBitHeight;
-    float           DigitalBitGap;
+    ImPlotSpec      Spec;
```
PeterJohnson, line 1195: I'm not that familiar with the internals, but what is ImPlotNextItemData's lifetime relative to the user-provided ImPlotSpec?  I ask because storing this by value forces a memcpy of the passed-in spec.  This is probably the safest thing to do, but does add a bit of overhead--storing a pointer would avoid the copy, but if lifetimes could differ (or user code could modify their ImPlotSpec in between uses of NextItemData), using a pointer would definitely be unsafe.

implot_internal.h
```diff
@@ -1173,30 +1192,17 @@ struct ImPlotNextPlotData
...
     ImPlotCond      HiddenCond;
     ImPlotNextItemData() { Reset(); }
     void Reset() {
-        for (int i = 0; i < 5; ++i)
-            Colors[i] = IMPLOT_AUTO_COL;
-        LineWeight    = MarkerSize = MarkerWeight = FillAlpha = ErrorBarSize = ErrorBarWeight = DigitalBitHeight = DigitalBitGap = IMPLOT_AUTO;
-        Marker        = IMPLOT_AUTO;
-        HasHidden     = Hidden = false;
+        Spec      = ImPlotSpec();
+        HasHidden = Hidden = false;
```
PeterJohnson, line 1205: Should HiddenCond be reset too?

PeterJohnson (commented):
I like the new API.  Just wanted to provide some code comments.

implot_internal.h
epezent: `ImPlotNextItemData` is a relic of the `SetNext` API (those functions would set variables on the current `ImPlotNextItemData `). I was tempted to remove it, but it's still doing a few useful things. Particularly, it stores transient flags about what should be rendered (e.g. `bool RenderLine` etc. which are established in `BeginItem`) and a transient `ImPlotSpec Spec`. This spec may or may not be exactly what the user passed! For instance, if the user-passed spec has any automatic variables, then `BeginItem` will convert them to whatever is appropriate and set `ImPlotNextItemData::Spec` accordingly. Thus, all the actual rendering code uses `ImPlotNextItemData::Spec` instead of the user provided spec. It's a little weird, admittedly, but it's functional.

The alternative, that I see, is that we pass specs to `PlotX` by non-const reference or pointer, update automatic variables in place, and use that throughout the render pipleine. Either of those would make the API less convenient than it currently is, and I think changing the user provided spec would cause confusion/issues.

PeterJohnson: Thanks, the current approach in this PR makes sense in that context.

implot_internal.h
epezent: Yes, it does. This seems to be an existing bug. Thanks for the catch!

ozlb:
Thanks for considering me!
I cross check and there is a bit of refactoring behind full implementation and I need to cross check a little bit more since i have some “ glitches “ that before was not present.
Most annoying issue for me so far, is that DigitalPadding and DigitalSpacing are now defined in style and so common for all plots.

ocornut:
I'm not equipped to provide overall feedback, but some small bits:

- Assuming this is a breaking change, I would suggest providing a couple of explicit example of "before" vs "after" code.
- The naming `ImProp` feels a bit too globally could it be ImPlotProp ? (if dear imgui ends up in a similar direction in the future we'd always have a change to merge the enums/types later)
- Good to spend a bit of time considering how non-C++ language wrapping would interact with the API (maybe there's no problem here?), and if there's be an decent path for e.g. [Dear Bindings](https://github.com/dearimgui/dear_bindings) to generate the corresponding metadata for implot.
- Wouldn't Designated Initializers solve Option 2 more naturally? (never used them in C++, just wondering)
- That's a personal committing guideline I use and you may not care about, but on large refactor I like to manually push all chunks for shallow parts (typo fixes, constexpr bits, comments) in a separate commit, so the big commit ends up as noise-less as possible. It also makes it easier to review PRs.

PeterJohnson:
> Most annoying issue for me so far, is that DigitalPadding and DigitalSpacing are now defined in style and so common for all plots.

Somehow I missed that on my first read-through.  That's a problem for me as well, as it sounds like this means it's no longer possible to arbitrarily vertically position individual digital plots on a per-line basis?

sonoro1234:
> With the default struct method, I think we are fine? @sonoro1234 what do you make of this?

Yes, I guess that with default struct method is ok.

PeterJohnson:
> I'm open to change if it if folks here don't mind the extra 4 characters, or if there's a more concise name anyone can think of.

I'm fine with a slightly longer name to avoid potential naming conflicts.

> Could spend more time on it if folks like it. Thoughts?  [Designated initializers] are a C++20 feature, unfortunately.

I personally wouldn't use either the string or pairs approach, as I'm using C++20. Strings sound not particularly performant.

Speaking of performance, one thing to note about large structs initialized through a default arguments (or via designated initializers) is that with many compilers this actually causes codegen to occur at *each* call site to fill out the struct, even though it's actually constant (the compiler won't generate a common constant struct in .rodata and just pass a pointer).  It looks like ImPlotSpec could get quite large and be passed as a default lots of places, so that may cause a surprising amount of codegen.  There's a complex reason for this (https://reddit.com/r/cpp/s/ezrziiv9NM has some discussion), but the workaround I found was to declare a `constexpr MyStruct kDefaultStruct;` and pass `kDefaultStruct` as the default argument value instead of `MyStruct()`.

> I don't use PlotDigital much myself, so let me know if this is problematic.

As long as it's still possible to set separately for each series via PushStyle, that works for me.

rokups:
Struct approach is good, this is what i would use as well. With variadic args approach and enums for parameters we still do not get hints about what type such parameter accepts unless we compile. With strings we do not get autocompletion for parameters too. Simplest approach is the best and that would be structs, with designated initializers, whenever we graduate to appropriately high standard.

ozlb:
> @ocornut
> 
> > The naming ImProp feels a bit too globally could it be ImPlotProp?
> 
> Yea, I was torn on this one for the same reasons. I leaned more toward `ImProp` since e.g. `ImPlotProp_LineColor` is a bit much for something that I intend people to use frequently. I'm open to change if it if folks here don't mind the extra 4 characters, or if there's a more concise name anyone can think of.

Probably both`ImAxis` and `ImProp` (and related enums) are now eligible to `ImPlot` prefix

ocornut:
There’s ImGuiAxis in the main codebase you could use that or decide its more consistent for users to see ImPlotAxis

PeterJohnson:
> I don't use PlotDigital much myself, so let me know if this is problematic. What I do want to avoid is adding digital plot specific variables to ImPlotSpec, for now (hence my decision to collapsing `DigitalBitHeight` into `ImPlotSpec::Size`).

It might make sense to create a ImPlotDigitalSpec struct to hold digital-specific settings?  The digital padding/spacing/height thing is kind of an oddity right now in that it affects behavior across multiple series and is done in pixel coordinates. I think spacing is really only useful if you're doing a lot of digital signals in one plot, and even then, the user could relatively easily compute this themselves?  It might be less confusing to just remove the padding feature entirely and just have a ImPlotDigitalSpec struct with a Y-axis start value?  This could also be a good time to think about whether pixel coordinates make sense for the start and height, or whether changing to something like % of plot height would be better.

A more generic solution overall could be https://github.com/epezent/implot/issues/321#issuecomment-1037072080 (in which case digital plots could always be full scale), but that major of a change is probably better done as a separate development effort.

marcizhu:
First of all, thanks for having me into consideration! And sorry for the super late reply, I've been meaning to reply to this for ages, but I haven't had the time until now.

I think the most important points have been raised already, but just to touch on a few things I'd like to point out that the second API feels more natural (kind of like what MATLAB, Python or other languages do, as you mentioned) while the first option is probably more performant, due to not having to iterate over the initializer list. Also, I haven't looked at the implementation for the first API, but if it takes in an rvalue reference it'd allow to use designated initializers, just like @ocornut mentioned, which is a big plus for people using C++20 and newer.

About the naming convention, I too would prefer `ImPlotProp` over `ImProp`, as the former is clearly part of ImPlot while the latter could perfectly be part of ImGUI or any other third-party library running on top of ImGUI. I have the strong opinion that nowadays, where everyone is using IDEs and code assistants and alike, a few more characters doesn't make a big difference since the autocomplete will still save you more characters than you have to type.

Another topic I see mentioned quite a few times is the digital plot and its style spec. Since the digital plot is completely different to a line/bar/pie chart, maybe it does make sense to have its style in a different struct. However, I have no strong opinions on this, so if this remains as-is it's fine by me. I still remember some conversations a while ago about doing some rework on the digital plot (issue #321 iirc) so I guess this change could also be made as part of that enhancement, since it would probably add tons of new styling options which might better justify having on its own struct.

That's pretty much all I wanted to say. Congrats on this PR! I think it's a nice step forward, and I'm very happy to see progress still being made on this awesome project! 😄

Nahor:
(I commented in the associated [discussion thread](https://github.com/epezent/implot/discussions/330) but I don't know how much visibility that discussion has)

What about using C++ types instead of an enum ([example](https://godbolt.org/z/eo6f895dT))? It's a bit more complicated/verbose on the implot side, but I think it's easier on the application side (less error prone, and more IDE friendly).

Nahor:
[Here is another option](https://godbolt.org/z/5djj87E18). It's similar to my previous proposal, but the "spec argument builder" are fully separated from the `ImPlotSpec`. This makes `ImPlotSpec` independent from the "builder functions" and allow an app to have its own "builder functions" if desired (e.g. to dynamically compute the color of a plot, or the line thickness), basically the equivalent of `ImPlotGetter` for `ImPlotSpec`. This also allows to have a generic "builder function" template and simplify the addition of new fields in `ImPlotSpec`.

Examples:
- Defining a new field using the generic template:
  ```C++
  typedef ImPlotSpec_Generic<ImVec4, offsetof(ImPlotSpec, LineColor)> ImPlotSpec_LineColor;
  ```
- Defining a custom builder function:
  ```C++
  struct ImPlotSpec_AutoLineWeight {
     static constexpr void assign(ImPlotSpec &spec, ImPlotSpec_LineWeight v) {
          spec.LineWeight = ImPlot::IsPlotHovered() ? 3.0f : 1.0f;
      }
  };
  ```
  which can be used like this:
  ```C++
  ImPlot::PlotLine(..., {ImPlotSpec_AutoLineWeight{}});
  ```
  which I think is nicer than:
  ```C++
  ImPlot::PlotLine(..., {ImProp_LineWeight, ComputeLineWeight()});
  ```

scanberg:
First, I'd like to thank @epezent for his ongoing work and dedication, which has made ImPlot so successful.
Second, as a user, I would like to offer my opinion on the design decision regarding styling mechanisms.

Option 1, using the pure struct approach, is the most straightforward approach to the problem.
As @ocornut points out, using designated initializers for setting field values is also a sufficient mechanism and does not introduce any C++ features that could hinder porting the functionality to other languages.

I would refactor it slightly to exclude data access parameters and flags to represent the pure styling parameters.
This could also be used as a solution to issue #332, which attempts to extend the Callback versions of plot procedures to allow for modifying the color based on the primitive's index.

I would approach the problem by having the user supply a styling struct upon invoking the plot procedure, then, in the callback, supply a reference to the style struct to be modified by the user if the user wants. This way, any relevant parameter such as color, size, or marker could be altered without having a combinatorial explosion of potential Getters.

```
struct ImPlotStyle {
    ImVec4          LineColor  = IMPLOT_AUTO_COL;       // line color (applies to lines, bar edges, marker edges); IMPLOT_AUTO_COL will use next Colormap color or current item color
    float           LineWeight = 1.0f;                  // line weight in pixels (applies to lines, bar edges, marker edges)
    ImVec4          FillColor  = IMPLOT_AUTO_COL;       // fill color (applies to shaded regions, bar faces, marker faces); IMPLOT_AUTO_COL will use next Colormap color or current item color
    float           FillAlpha  = 1.0f;                  // alpha multiplier (applies to FillColor)
    ImPlotMarker    Marker     = ImPlotMarker_None;     // marker type; specify ImPlotMarker_Auto to use the next unused marker
    float           Size       = 4;                     // size of markers (radius), error bar whiskers (width or height), and digital bars (height) *in pixels*
     ...
}

ImVec2 pos[20] = {...};

struct Data {
    const ImVec2* pos;
    ImPlotColormap map;
};

Data data = {
    .pos = pos,
    .map = ImPlotColormap_Plasma,
};

ImPlotLineFlags flags = 0;
ImPlotStyle style = {
    .Size = 10.f,
};

// Render using style
ImPlot::PlotLine("Line 1", &data.pos[0].x, &data.pos[0].y, 20, flags, 0, sizeof(Data), style);

// Example with getter
auto getter = [](int idx, ImPlotStyle& style, void* user_data) -> ImPlotPoint {
    Data* data = (Data*)user_data;
    style.LineColor= ImPlot::GetColormapColor(idx, data->map);
    return {data->pos[idx].x, data->pos[idx].y};
};

// Plot lines using getter
ImPlot::PlotLineG("Line 2", getter, data, 20, flags, style);
...
```
In this design, passing the style parameter as const ImPlotStyle& makes sense, which is set to {} by default.
Then, in the case of PlotLineG, it is internally copied to allow for modification when iterating over items.

keltonhalbert:
I'd really love to know @epezent's take on whether more time has provided clarity on which direction to go, and where #332 fits into this. Styling lines and scatters with different sizes and colors would be a really excellent and powerful feature, especially for something like ImPlot/ImGui and its robust support for realtime interactivity. 

I just finished merging the changes from the #332 PR into a local branch of mine so that I can go ahead and use this functionality, but I would really love to contribute functionality to ImPlot in any way I can. Unfortunately, without clear guidance from maintainers such as @epezent / @ocornut, I'm a little apprehensive to go too far down the wrong path only to have a different solution/direction adopted by the main repository.

My main goal is to use this feature to color one line based on the value of another. The approach from #332 has worked so far, coloring the line in the upper right panel based on the distance between the white and red lines in the upper left plot. This is in an effort to write a portable, interactive viewer for weather data used by the NOAA Storm Prediction Center. If in my efforts to write this application, I can provide testing, source code contributions, and ideas, I'd really like to help push this forward if possible -- just point me in the right direction!
image

epezent:
@keltonhalbert Thank you for reaching out. I commented on the status of ImPlot recently in https://github.com/epezent/implot/issues/585#issuecomment-2270021862 and the situation remains the same -- I am not actively contributing or addressing PRs at this time, but I am openly welcoming any serious contributors/maintainers to take lead. If you are interested in discussing this, please do not hesitate to reach out via email -- Evan

keltonhalbert:
> @keltonhalbert Thank you for reaching out. I commented on the status of ImPlot recently in [#585 (comment)](https://github.com/epezent/implot/issues/585#issuecomment-2270021862) and the situation remains the same -- I am not actively contributing or addressing PRs at this time, but I am openly welcoming any serious contributors/maintainers to take lead. If you are interested in discussing this, please do not hesitate to reach out via email -- Evan

Thanks Evan! I'll absolutely be reaching out to you... I've only been spinning myself up on ImGui/ImPlot over the last few months, but I am very interested and invested in seeing this library continue to progress forward.

brenocq:
I've just rebased this branch on top of `master` to get the ball rolling again. `ImPlotSpec` can now be easily tested by running the example:

```bash
cd implot
cmake -B build && cmake --build build && build/example
```
Since the rebase was significant, I may have missed something. Please let me know if you encounter any bugs or have fresh thoughts on the implementation.

After discussing with @epezent, the plan is to release this feature in **v0.18**, so I'd love to revive this thread :)

cc: @PeterJohnson @ozlb @ocornut @sonoro1234 @rokups @marcizhu @scanberg @keltonhalbert

brenocq:
The current implementation of `ImPlotProp` uses the same `LineColor` for both marker outline and line color. This can prevent the user from setting different colors for the line and for the markers (they would need to call the plot function twice). I'll follow the same standard as matplotlib.

```python
plt.plot(x, y, 
         color='blue',           # Color of the line
         marker='o',             # Type of marker
         markerfacecolor='red',  # Color of the marker's inside
         markeredgecolor='green',# Color of the marker's border
         markersize=10)
```

I'll add `MarkerLineColor` and `MarkerFillColor` to `ImPlotProp`. By default, they'll be set to `LineColor`, having the same result as before if the user doesn't want a different color for the markers.

sonoro1234:
Having problems with parsing implot_internal:

```cpp
template<typename TContainer> struct value_type { using type = typename TContainer::value_type; };
```

brenocq:
> Having problems with parsing implot_internal

Thanks for testing it! I'll try to revert these changes to before

sonoro1234:
I dont know if dropping internal generation is so bad?!

brenocq:
> I dont know if dropping internal generation is so bad?!

What do you mean here? Removing all templates from internal?
I'm trying to rewrite it to avoid the fancy template value resolution.

brenocq:
@sonoro1234 I reverted most of the changes from the template machinery. It's not great that now there's some duplicated code in `implot_items.cpp`, but hopefully it should help your parser. Let me know if you have any other issues.

brenocq:
After renaming `imProp` to `ImPlotProp`, it got quite verbose to set the properties
```cpp
ImPlot::PlotScatter("Data 2", xs2, ys2, 50, {
    ImPlotProp_Marker, ImPlotMarker_Square,
    ImPlotProp_Size, 6,
    ImPlotProp_LineColor, GetColormapColor(1),
    ImPlotProp_FillColor, GetColormapColor(1),
    ImPlotProp_FillAlpha, 0.25f
});
```

I'll bring back the string style @epezent commented in https://github.com/epezent/implot/pull/519#issuecomment-1745484398. While we can't do compile time checks when the user chooses this style, at least we can still have runtime assertions. 
```cpp
ImPlot::PlotScatter("Data 2", xs2, ys2, 50, {
    "Marker", ImPlotMarker_Square,
    "Size", 6,
    "LineColor", GetColormapColor(1),
    "FillColor", GetColormapColor(1),
    "FillAlpha", 0.25f
});
```

sonoro1234:
> What do you mean here? Removing all templates from internal?
> I'm trying to rewrite it to avoid the fancy template value resolution.

Not at all, I mean that I could skip generation comming from implot_internal.h as it does not seem so necessary. (Now I was using implot.h and also implot_internal.h)

After you last changes I can use imgui_internal.h without problems.

brenocq:
I've created a new property `MarkerSize`. The previous generic `Size` property is now only used for error bar whiskers and digital plot height

brenocq:
I believe this PR is in a good state to be merged. Let me know if you guys have any issue or suggestions @PeterJohnson @ozlb @ocornut @sonoro1234 @rokups @marcizhu @scanberg @keltonhalbert

ocornut:
Very casual check:

(1) This add `#include "imgui_internal.h"` to implot.h, seems like a problem to me we should address.

(2) I feel a bit uneasy about the version that takes string for identifier. 
  - It feels a bit magical and unlike ImGui style practices. Also unlike other ImPlot features.
  - It is arguably and definitively slower. This library ecosystem works because it strives to be optimized everywhere.
  - It would encourage high-level language bindings to adopt a similar style, and god-knows how that would be implemented.
  - Promoting 3 ways of doing a same thing seems unnecessary and can make things more confusing to users (think: core snippet sharing).
  - It makes compile-checks harder.
  - It prevents IDE for using completion.
  - ....all to save typing 9 characters. Really not worth it IMHO.

(3) I recommend always incrementing `IMPLOT_VERSION_NUM` when pushing non-trivial changes, even before a release is made (well, particulary before a release is made).

ocornut:
(4) in the spirit of imgui.h / imgui.cpp "API BREAKING CHANGES" sections, I think you should keep old symbol commented out with help about how to proceed.

So you would turn e.g.
```cpp
// Set the line color and weight for the next item only.
IMPLOT_API void SetNextLineStyle(const ImVec4& col = IMPLOT_AUTO_COL, float weight = IMPLOT_AUTO);
```

Into a comment in the `// [SECTION] Obsolete API` section:

```
//void SetNextLineStyle(const ImVec4& col = IMPLOT_AUTO_COL, float weight = IMPLOT_AUTO); // OBSOLETED IN 0.18 // Set ImPlotSpec.LineColor/LineWeight or construct ImPlotSpec with { ImPlotSpec_LineColor, color }, { ImPlotSpec _LineWeight, weight }.
```

brenocq:
Thank you so much for the detailed feedback @ocornut! I've just pushed the changes for (1), (2), (3), (4)

sonoro1234:
Everything works!!
Observation: Taking ImPlotSpec instead of ImPlotSpec* makes C wrapper to declare ImPlotSpec_c as the C counterpart of a non UDT struct(with constructor for example, see:https://github.com/cimgui/cimgui/issues/298 and https://github.com/cimgui/cimgui/issues/309). The problem is then initialize ImPlotSpec_c. If ImPlotSpec* was used then initialization would be trivial by calling ImPlotSpec constructor.

This is working now:

```c
ImPlotSpec spec = *ImPlotSpec_ImPlotSpec();
...
ImPlot_PlotBars_S64PtrInt("My Bar Plot", bar_data, 11,0.67,0,spec);
```

but it seems that could fail in ARM64 (I cant verify it myself)

brenocq:
Since there's no more feedback, I'll merge it into the `master` now. Do let me know if you guys experience any problems.

epezent:
I am so happy to see this FINALLY merged. Awesome work @brenocq!

Thanks for the input @ocornut @sonoro1234

epezent:
> Very casual check:
> 
> (1) This add `#include "imgui_internal.h"` to implot.h, seems like a problem to me we should address.
> 
> (2) I feel a bit uneasy about the version that takes string for identifier.
> 
> * It feels a bit magical and unlike ImGui style practices. Also unlike other ImPlot features.
> * It is arguably and definitively slower. This library ecosystem works because it strives to be optimized everywhere.
> * It would encourage high-level language bindings to adopt a similar style, and god-knows how that would be implemented.
> * Promoting 3 ways of doing a same thing seems unnecessary and can make things more confusing to users (think: core snippet sharing).
> * It makes compile-checks harder.
> * It prevents IDE for using completion.
> * ....all to save typing 9 characters. Really not worth it IMHO.
> 
> (3) I recommend always incrementing `IMPLOT_VERSION_NUM` when pushing non-trivial changes, even before a release is made (well, particulary before a release is made).

Agree with @ocornut on (2). The string version cons far out weigh having to type 9 extra characters. I implemented this path out of curiosity and novelty. In hindsight, it's just not a good idea.

pthom:
@brenocq : (This is FYI only). I'm updating the python bindings in imgui_bundle following this PR.

I cannot bind the variadic template constructor `ImPlotSpec(Args... args)`, and the templated SetProp (`template <typename Arg, typename ...Args>
    void SetProp(ImPlotProp prop, Arg arg, Args... args)`), because those are very much C++ specific.


However, I am lucky because my bindings generator will generate a [constructor with named parameters](https://github.com/pthom/imgui_bundle/blob/c3b86b10e0372bf241169c4083585c9e2bf37c63/bindings/imgui_bundle/implot/__init__.pyi#L925-L952) for ImPlotSpec.
    
... So that in Python we can write:

```python
        implot.plot_line("line", values,
                        spec=implot.Spec(
                            line_color=color,
                            fill_color=color,
                            fill_alpha=0.25,
                            offset=offset,
                            flags=implot.LineFlags_.shaded))
```

I think this should be enough for Python: keyword arguments are quite clear in this particular case (and of course, any order is acceptable)

pthom:
@brenocq : 
I will soon release imgui_bundle v1.92.6. It will probably include ImPlot "0.18 WIP" (i.e. with this PR included).

pthom:
We have the same in the python API for implot3d:
https://github.com/pthom/imgui_bundle/commit/e15e360733e5e7c508794b433086fdb7972365b1

PR #549, opened 2023-12-29
Add IMPLOT_API to constructors and functions of ImPlotPoint, ImPlotRange, and ImPlotRect (fixes #545)

### LINKED ISSUE #545: Constructors and Function of structs don't have the IMPLOT_API macro
I've been using the https://github.com/aybe/DearImGui project to generate C# bindings for imgui and implot, and I ran into an issue (https://github.com/aybe/DearImGui/issues/24) where calling the constructor for ImPlotPoint would cause an EntryPointNotFoundException.

After digging into it some, the root issue is simply that none of the constructors for ImPlotPoint wind up in the dll's exports when using the IMPLOT_API macro when that expands to "__declspec(dllexport)". In fact, none of those constructors - and for that matter, none of the constructors or member functions of any other struct in implot - have the IMPLOT_API macro in front of them in the source.

Was this an intentional choice in implot?

I checked with imgui's source and the structs there (e.g. ImGuiStyle) do have IMGUI_API declared for constructors and member functions.

Cidolfas:
This is to allow an exported DLL to access these functions, e.g. so they can be called in C#

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -467,11 +467,11 @@ enum ImPlotBin_ {
 IM_MSVC_RUNTIME_CHECKS_OFF
 struct ImPlotPoint {
     double x, y;
-    constexpr ImPlotPoint()                     : x(0.0), y(0.0) { }
-    constexpr ImPlotPoint(double _x, double _y) : x(_x), y(_y) { }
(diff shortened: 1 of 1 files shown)


brenocq:
Thanks for the contribution @Cidolfas! I've also added the `IMPLOT_API` to the locator functions.

Issue #256, opened 2021-06-22
ShowDatePicker Misses May, June, July, August, September, October


peterwilson136:
Using the code below, if you click on `June 2021` -> `Jan` ->`1` (and make sure you click on the `1`), and then you go back to `January 2021` -> `Jun`, it puts the date to `31 May, 2021`. You can repeat this for the dates listed in the title, and you will see that it only works up till `April` and after `November`. 
```
#include <GL/glew.h>

#ifndef GLFW_INCLUDE_NONE
#define GLFW_INCLUDE_NONE // GLFW including OpenGL headers causes ambiguity or multiple definition errors.
#endif // GLFW_INCLUDE_NONE

#include <GLFW/glfw3.h>

#include "ImGui/imgui.h"
#include "ImGui/implot.h"
#include "ImGui/implot_internal.h"
#include "ImGui/imgui_impl_glfw.h"
#include "ImGui/imgui_impl_opengl3.h"


int main(int argc, char* argv[])
{
    glfwInit();

    const char* glslVersion = "#version 460";
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 6);
    GLFWwindow* window = glfwCreateWindow(1280, 720, "Program", NULL, NULL);
    glfwMakeContextCurrent(window);
    glfwSwapInterval(0); //  vsync

    glewInit();

    IMGUI_CHECKVERSION();
    ImGui::CreateContext();
    ImPlot::CreateContext();

    ImGui::StyleColorsDark();
    ImPlotStyle& ImPlot_Style = ImPlot::GetStyle();
    ImPlot_Style.UseLocalTime = true;
    ImPlot_Style.AntiAliasedLines = true;

    ImGui_ImplGlfw_InitForOpenGL(window, true);
    ImGui_ImplOpenGL3_Init(glslVersion);

    static ImPlotTime from = 1623938400 - 86400;
    static int fromLevel = 0;

    while (!glfwWindowShouldClose(window))
    {
        glfwPollEvents();
        ImGui_ImplOpenGL3_NewFrame();
        ImGui_ImplGlfw_NewFrame();
        ImGui::NewFrame();

        ImGui::Text("%.3f", ImGui::GetIO().Framerate);
        ImPlot::ShowDatePicker("##ComparisonBefore", &fromLevel, &from, nullptr, &from);
    
        ImGui::Render();
[...]

PR #568, opened 2024-05-11
handle possible DST change within FloorTime()

shrddr:
see https://github.com/epezent/implot/issues/256


shrddr (commented):
According to https://stackoverflow.com/questions/8558919/mktime-and-tm-isdst `gp.Tm.tm_isdst = -1;` can sometimes produce unexpected result - when the return value of `FloorTime()` is within that one hour in October/November when same Local time has 2 possible GMT interpretations. When calling `FloorTime(unit=year/month/day)`, it is not a problem, but might potentially become a problem when called with `unit==hour/minute/second`.

shrddr:
Additionally made major ticks nicer looking

before
image

after
image

PR #570, opened 2024-05-17
replace older time functions with more portable ones

Gargaj:
As per https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -909,7 +909,7 @@ ImPlotTime MkGmtTime(struct tm *ptm) {
 #ifdef _WIN32
     t.S = _mkgmtime(ptm);
 #else
-    t.S = timegm(ptm);
+    t.S = mktime(ptm);
 #endif
     if (t.S < 0)
         t.S = 0;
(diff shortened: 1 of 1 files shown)


epezent:
@Gargaj I think your code is missing an important step:

```
Portability note: mktime is essentially universally available. timegm is rather rare. For the most portable conversion from a UTC broken-down time to a simple time, set the TZ environment variable to UTC, call mktime, then set TZ back.
```

Gargaj:
This has been a while so I don't specifically remember what was being broken, but the functions I was trying to fix were not available on one of the current generation consoles.

skarolewics:
I am encountering the same issue as @Gargaj, and the platform in question does not support time zone setting via functions like `tzset()`.

I was able to workaround by defining replacements for `timgm` and other functions before including implot.cpp, but it is a bit of a hack. Perhaps for robustness of the ImPlot API, a `#define` could be optionally provided to override implementation of the time related functions, to enable implementation sites to control behavior for specific platforms.

CA-Jon-Colverson:
@skarolewics Would you be able to share your replacement for `timegm`?

skarolewics:
@CA-Jon-Colverson I used the solution described here: https://stackoverflow.com/a/68761107

PR #591, opened 2024-09-23
feat: add `IsPlotChanging()`

SamuMazzi:
As pointed out in the `TODO.md`, this PR implements a function to detect if the limits of the plots are changing.


epezent:
@SamuMazzi Thanks for taking a swing at this. A couple things:

I think the following code in SetupAxis will only be called every frame if the user explicitly calls SetupAxis/SetupAxes in their plot code block. Since this is not required, this feature will not work in some scenarios (e.g. the `Plot Shaded` demo and others)

```
// cache previous range
axis.PreviousRange = ImPlotRange(axis.Range.Min, axis.Range.Max);
```

I think the solution here is to cache the previous range(s) in SetupFinish, just before UpdateInput. Lemme know what you think.

Second, I suggest we name this function something more descriptive and also consider if it should be callable per-axis, for instance:

```
bool IsAxisRangeChanging(ImAxis axis)
```

SamuMazzi:
> Since this is not required, this feature will not work in some scenarios (e.g. the Plot Shaded demo and others)

Nice catch, thank you, I didn't think about that!

> Second, I suggest we name this function something more descriptive and also consider if it should be callable per-axis

Yes, perfect! I used the previous name just because of the notes in the `TODO.md`, but it makes more sense this name (and also the selection of the axis) ;)

And also sorry, I made a little mess with commits, hope it's not a problem!

PR #592, opened 2024-09-23
feat: add `ImPlotPieChartFlags_Exploding`

SamuMazzi:
This PR add the `ImPlotPieChartFlags_Exploding` flag to the pie chart so that when you hover the legend, the corresponding item in the chart explodes


epezent:
Nice work! Merged

PR #608, opened 2025-01-18
Custom per-item colors for line and scatter plot

mihaly-sisak:
Hello!
I've implemented user-settable per-item colors for line and scatter plots.
I've tried to match your coding style as close as I was able.
Please consider merging this pull request, or provide feedback how can I improve the implementation so it can be merged.
All previously available public plot APIs are unchanged.
I've added two new plot types:
```
typedef ImU32 (*ImPlotColorGetter)(ImPlotCol col, int idx, void* user_data);
IMPLOT_API void PlotLineCG(const char* label_id, ImPlotGetter getter_func, void* data, int count, ImPlotColorGetter color_func, void* color_data, ImPlotLineFlags flags=0);
IMPLOT_API void PlotScatterCG(const char* label_id, ImPlotGetter getter, void* data, int count, ImPlotColorGetter color_func, void* color_data, ImPlotScatterFlags flags=0);
``` 
The user can write a callback to set the item color based on the index.
I've added to the demo to showcase the new APIs.
The single-color plots are drawn with a new `struct ColorGetter` that is basically just a wrapper around `ImGui::GetColorU32(GetItemData().Colors[col])`.
I wanted to cache the plot colors so the ImVec4->ImU32 conversion happens only once. If you can give me any direction how can I achieve this more elegantly (than the current `fill_colors()`) please point me to the right direction. With this setup the ImVec4->ImU32 conversion happens only once, just as before with the single-color only version.
The per-item-colored plots are drawn with `struct ColorGetterFuncPtr`, implemented in the same style as `struct GetterXY` and `struct GetterFuncPtr`.

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -861,6 +861,7 @@ IMPLOT_API void SetNextAxesToFit();
 IMPLOT_TMP void PlotLine(const char* label_id, const T* values, int count, double xscale=1, double xstart=0, ImPlotLineFlags flags=0, int offset=0, int stride=sizeof(T));
 IMPLOT_TMP void PlotLine(const char* label_id, const T* xs, const T* ys, int count, ImPlotLineFlags flags=0, int offset=0, int stride=sizeof(T));
 IMPLOT_API void PlotLineG(const char* label_id, ImPlotGetter getter, void* data, int count, ImPlotLineFlags flags=0);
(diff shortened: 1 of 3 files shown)


mihaly-sisak:
Renamed `fill_colors()` to `FillColors()` I dont know how I did not realize the proper naming convention.

brenocq:
Hi @mihaly-sisak! I'm closing this PR since #672 was merged. Thank you so much for the initial effort, your code served as a lot of inspiration while implementing #672 😃

PR #611, opened 2025-02-19
Added PlotBubbles 

Luc16:
**Summary**  
This PR adds support for bubble plots to ImPlot, inspired by [MATLAB’s bubblechart](https://www.mathworks.com/help/matlab/ref/bubblechart.html). Existing plot functionality is unaffected.

---

**Function signatures** 

```
IMPLOT_TMP void PlotBubbles(const char* label_id, const T* values, const T* szs, int count, double xscale=1, double xstart=0, ImPlotBubblesFlags flags=0, float min_pxsize=3, float max_pxsize=50, int offset=0, int stride=sizeof(T));

IMPLOT_TMP void PlotBubbles(const char* label_id, const T* xs, const T* ys, const T* szs, int count, ImPlotBubblesFlags flags=0, float min_pxsize=3, float max_pxsize=50, int offset=0, int stride=sizeof(T));

IMPLOT_API void PlotBubblesG(const char* label_id, ImPlotGetter3D getter, void* data, int count, ImPlotBubblesFlags flags=0, float min_pxsize=3, float max_pxsize=50);
```

- min_pxsize and max_pxsize set the minimum and maximum bubble diameters (in pixels).
- By default, marker outlines are not rendered and the fill color alpha is set to 0.5.

---

**Example**
Below is an example bubble plot (two different datasets). The bubble sizes are scaled between the specified min_pxsize and max_pxsize ([3, 50] for blue and [5, 20] for orange):

image

---

**Feedback**
Please let me know if there are any suggestions or improvements regarding:

- The API signature or naming.
- Default styling choices (e.g., alpha, line rendering).
- Performance or edge-case handling.

I appreciate any comments or reviews on how to refine this feature!

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -13,7 +13,7 @@ The list below represents a combination of high-priority work, nice-to-have feat
 
 ## Plot Items
 
-- add `PlotBubbles` (see MATLAB bubble chart)
+- add legend for `PlotBubbles` (see MATLAB bubble chart)
 - add non-zero references for `PlotBars` etc.
 - fix appearance of `PlotBars` spacing
 
--- a/implot.h
+++ b/implot.h
@@ -885,9 +885,9 @@ IMPLOT_TMP void PlotScatter(const char* label_id, const T* xs, const T* ys, int 
 IMPLOT_API void PlotScatterG(const char* label_id, ImPlotGetter getter, void* data, int count, ImPlotScatterFlags flags=0);
 
 // Plots a bubble graph. Default marker is ImPlotMarker_Circle.
-IMPLOT_TMP void PlotBubbles(const char* label_id, const T* values, const T* szs, int count, double xscale=1, double xstart=0, ImPlotBubblesFlags flags=0, int offset=0, int stride=sizeof(T));
-IMPLOT_TMP void PlotBubbles(const char* label_id, const T* xs, const T* ys, const T* szs, int count, ImPlotBubblesFlags flags=0, int offset=0, int stride=sizeof(T));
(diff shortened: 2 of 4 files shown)


brenocq:
Thank you for taking the time to implement The `PlotBubbles` @Luc16! I've just tried it here:

https://github.com/user-attachments/assets/616c4e94-1d57-4d1a-9837-d525bb143b7a

There are some things that I would suggest to change:
(1) The bubbles should still be rendered when they are partially seen. Right now they stop being rendered when the center is outside the plotting range.
(2) I believe it would be better to have the bubble size in plot units so they are scaled together with the zoom in/out (the PlotBars also has size in plot units)
(3) I think it's better to remove the `min_pxsize` and `max_pxsize` to simplify the API
(4) In the future, PlotScatter should support plotting markers with different sizes, so I think it would be better to hardcode PlotBubbles to plot circles (with number of vertices depending on circle size in pixels). The circle edge color could be defined by the LineColor and the circle fill by the FillColor and FillAlpha.

I'll start working on these changes in your branch, do let me know before you start working on any change so we don't do duplicated work

brenocq:
I've just finished applying my suggestions, I created a new `RendererCircleFill` and `RendererCircleLine` to handle the circle rendering. These renderers vary the number of segments to be rendered depending on the circle size.

https://github.com/user-attachments/assets/9e6151a4-7cd4-4717-9803-dda3024ae2d6

Luc16:
Looks great to me.

I agree with the changes and the API simplification. The new circle renderers and plot-unit sizing are definitely better aligned with ImPlot’s design.

I’m happy to merge this as is. Thanks for the refinements :)

PR #613, opened 2025-02-20
Add custom right-click context menu

### LINKED ISSUE #530: Add items to context menu
Is it possible to define additional options from the right-click menu?
I looked through the code and I can't find a solution other than modifying the ImPlot code.
I think it would be a useful feature to allow us to expose extra functionality to the users.

moe93:
Closes #530

Usage:

```
if (ImPlot::BeginPlot("My Plot", ImVec2(-1, 0), ImPlotFlags_NoCentralMenu))
{
    // Plot things before

    if (ImPlot::BeginCustomContext())
    {
        if (ImGui::MenuItem("My Custom Item")) doThing();
        ImPlot::EndCustomContext(true); // true = append standard menu
    }
    ImPlot::EndPlot();

    // Plot things after
}
```

This is the work of @PapaNaxos. I just packaged it and created the PR.


PapaNaxos:
I originally made this comment on the associated issue, but maybe it's better to put here.

> I do wonder if `ImPlotFlags_NoCentralMenu` is the most intuitive flag name or not... 
> 
> Even though I wrote it, I don't find it particularly obvious that THAT flag is what is required to use a custom context menu 😅 
> 
> Perhaps
>   `ImPlotFlags_HasCustomContext`  
>   `ImPlotFlags_UseCustomContext`  
>   `ImPlotFlags_CustomContext`  
> ?

PapaNaxos:
Does 'assigned' mean I need to do... something?

brenocq:
> Does 'assigned' mean I need to do... something?

Huehuehe if I leave comments while reviewing you might want to do... something :)

Actually, I don't think you can modify the code in this PR, right? Damn GitHub

brenocq:
I'm trying to think on how to make the custom context menus more flexible. Right now, depending on where the user clicks (plot, axis, legend), a different context menu is shown. So I would suggest allowing the user specify which context menu they want to customize. Additionally, we should expose the `ShowDefault*ContextMenu` to the user so they can keep the default context menu content if wanted. It should look something like this:

```cpp
if (ImPlot::BeginPlot("My Plot", ImVec2(-1, 0)))
{
    // Plot things before

    if (ImPlot::BeginPlotContextMenu())
    {
        if (ImGui::MenuItem("My Custom Item")) doThing();
        ImGui::Separator();
        ImPlot::ShowDefaultPlotContextMenu();
        ImPlot::EndPlotContextMenu();
    }
    
    if (ImPlot::BeginAxisContextMenu(ImAxis_X1))
    {
        ImGui::Text("Overwritten X Axis Context Menu");
        ImPlot::EndAxisContextMenu();
    }

    if (ImPlot::BeginLegendContextMenu())
    {
        ImGui::Text("My Legend Items:");
        ImPlot::ShowDefaultLegendContextMenu();
        ImPlot::EndLegendContextMenu();
    }
    
    ImPlot::EndPlot();
}
```

I don't really love that it is a bit verbose, but it makes the behavior quite clear (hopefully). Let me know what you guys think! @moe93 @PapaNaxos

moe93:
I am of the opinion that having context menus appear based on what item is clicked is more practical.
It does make the intent more clear and allows for a more intuitive behavior in terms of the end-user.

For instance, when plotting vibration data, I allow the users to choose whether to display the x-axis in RPMs, Hz, or seconds. Their biggest gripe was that they expected to find those choices when right-clicking the x-axis, and NOT within the custom context menu found by right-clicking the plot area.
I wholeheartedly agree with their sentiment.

PapaNaxos:
That's a far better interface I think!

WhatWhatz:
Any updates on when this will be merged/modified?

PR #626, opened 2025-06-26
Add row and column offset as well as row and column stride for the heatmaps

### LINKED ISSUE #625: Feature: Add row and column offset for heatmaps
Thank you for the amazing library. 

I have a use case where I store and add data in a circular buffer which is used to draw a heatmap. The circular buffer is preallocated and once it is full we start replacing data from the start. 
To avoid having to shift and copy around data I suggest we add a column and row offset to the heatmap function which will shift the rows and columns based on the passed in values

ACvanWyk:
Fixes #625 
Added implementation for specifying a row and column offset as well as row and column stride for the heatmap plot.
This will shift the rows and the columns of the heatmap based on the passed in offset and sweep values. By default there will not be a offset or step added to the rows or columns.  This should work for both row major and column major plots. 
I added a little demo, `Demo_HeatmapOffsets`, of how it would work to add row and column offsets + row and column stride to the demo application.
Could potentially add the same parameters to other plots that use the heatmap under the hood like the Histogram 2D. 
I tried implementing it is such a way that does not break the existing API but I might have missed something.


brenocq:
I'll be closing this issue since #519 added support for `PlotHeatmap` offset and stride. See https://github.com/epezent/implot/issues/625#issuecomment-3912370201.

PR #661, opened 2025-12-03
fix: `DragRect` resizing when its size is zero

JunkyoLee:
Previously, DragRect could not be resized if its current size was zero, which prevented users from initializing or adjusting the rectangle from a zero-size state.

Modified the logic to allow moving the rect only when either its width or height is non-zero


brenocq (approved):
Lovely!!

`master`:

https://github.com/user-attachments/assets/ad28f59c-ef01-4544-b1fb-80a9c8aa4c6e

`dragrect_resizing`:

https://github.com/user-attachments/assets/47c0156f-919c-4f3b-9de3-4ee8151d84f9


Thanks for the minimal changes and for following conventional commits :)

PR #665, opened 2025-12-31
Feature: Implementation of Per Point Custom Color for Line, Scatter and nd Contour

vincentjzy:
This PR provides the function that users can add an array storing the value at each point (such as temperature, density, etc.) and assign the color to each point according to the selected colormap. The existing functionality and invoking methods are not influenced by this modification. The demo can be found in the figures below. We also made similar implementation for ImPlot3D (https://github.com/brenocq/implot3d/pull/162) .

PerPointCustomColor_Lines
PerPointCustomColor_Scatters
PerPointCustomColor_Contour

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -4596,6 +4596,19 @@ ImVec4 SampleColormap(float t, ImPlotColormap cmap) {
     return ImGui::ColorConvertU32ToFloat4(SampleColormapU32(t,cmap));
 }
 
+ImU32 ConvertValueToColor(float value, float v_min, float v_max, ImPlotColormap colormap) {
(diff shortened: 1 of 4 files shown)


brenocq:
Hi @vincentjzy! Since `ImPlotSpec` (#519) is now merged into main, I reimplement this per point custom colors in a new PR #672 by following the `ImPlotSpec` style. I would love some feedback on #672!

I still need to implement something like #672 for ImPlot3D

brenocq:
Hi @vincentjzy! I'll close this PR since this feature was implemented in #672. Thank you so much for the initial effort, I drew a lot of inspiration from your code when implementing on #672.

PR #673, opened 2026-02-14
Feat: PlotPolygon

### LINKED ISSUE #631: [Feature] Plot polygon
Hi, love this library, thank you for all the great work.  I'm trying to use the fill features to add some shading within plotted shapes and noticing that the fill doesn't seem to be as precise for the concave shapes I'm drawing.

See this image:

Image

The red line should be the boundary of the shape, and I was hoping to fill within the boundary.  You can see that the fill is spilling out from the shape to make it more square (and it seems to be a different opacity). 

Currently I'm drawing the outline as a clockwise loop.  The fill is drawn using plotShaded with the exact same clockwise loop of points.

I was wondering if this is user error or a known issue or if there were any work arounds here?  I could tesselate this shape by hand, 

Thanks again!

Edit, also just noticed that the fill on the top doesn't line up either:

Image

brenocq:
Closes #631

This PR introduces `PlotPolygon` as requested by #631.

```cpp
// Flags for PlotPolygon. Used by setting ImPlotSpec::Flags.
enum ImPlotPolygonFlags_ {
  ImPlotPolygonFlags_None     = 0,       // default (closed, convex polygon)
  ImPlotPolygonFlags_Concave  = 1 << 10, // use concave polygon filling (slower but supports concave shapes)
};

// Plots a polygon. Points are specified as separate x and y arrays. Supports both convex and concave polygons.
IMPLOT_TMP void PlotPolygon(const char* label_id, const T* xs, const T* ys, int count, const ImPlotSpec& spec=ImPlotSpec());
```

**Usage**

```cpp
...
if (ImPlot::BeginPlot("Polygon Plot", ImVec2(-1,0), ImPlotFlags_Equal)) {
    ImPlot::PlotPolygon("Triangle", tri_xs, tri_ys, 3, {
        ImPlotProp_FillAlpha, 0.5f,
    });
    ImPlot::PlotPolygon("Pentagon", pent_xs, pent_ys, 5, {
        ImPlotProp_FillAlpha, 0.5f,
        ImPlotProp_FillColor, ImVec4(0,1,0,1),
    });
    ImPlot::PlotPolygon("Star (Concave)", star_xs, star_ys, 10, {
        ImPlotProp_FillAlpha, 0.5f,
        ImPlotProp_FillColor, ImVec4(1,1,0,1),
        ImPlotProp_Flags, ImPlotPolygonFlags_Concave,
    });

    ImPlot::EndPlot();
}
```
image

### DIFF
--- a/implot_items.cpp
+++ b/implot_items.cpp
@@ -1965,7 +1965,7 @@ void PlotPolygonEx(const char* label_id, const Getter& getter, const ImPlotSpec&
         bool flip = !((x_inv ? 1 : 0) ^ (y_inv ? 1 : 0));
 
         // Transform all points to screen space
-        ImVec2* points = (ImVec2*)alloca(getter.Count * sizeof(ImVec2));
+        ImVec2* points = (ImVec2*)IM_ALLOC(getter.Count * sizeof(ImVec2));
         for (int i = 0; i < getter.Count; ++i) {
             ImPlotPoint p = flip ? getter[getter.Count - 1 - i] : getter[i];
             points[i] = transformer(p);
@@ -1982,6 +1982,7 @@ void PlotPolygonEx(const char* label_id, const Getter& getter, const ImPlotSpec&
             const ImU32 col_line = ImGui::GetColorU32(s.Spec.LineColor);
             draw_list.AddPolyline(points, getter.Count, col_line, ImDrawFlags_Closed, s.Spec.LineWeight);
         }
+        IM_FREE(points);
 
         EndItem();
     }


brenocq:
Since the user specifies the points in plot space, but the ImGui draw list methods expect the points in screen space, I needed dynamically allocate an array of ImVec2 to convert each point from plot space to screen space before sending the array to ImGui, I think this is fine for now for an initial implementation, in the future we can create a polyline renderer so we avoid this dynamic allocation by filling in the draw list directly.

Issue #667, opened 2026-01-07
[Build] : CMake Build Support


djoezeke:
Dear ImPlot Community...

I'm proposing CMake support to ImPlot.

brmarkus:
Would you mind describing your use-case, please?
Is it due to the dependencies? Is it that vcpkg is not available? Is it manually adding the `implot.h, implot_internal.h, implot.cpp, implot_items.cpp and optionally implot_demo.cpp` and you want to get them downloaded as part of cmake config step?

djoezeke:
> Would you mind describing your use-case, please?
> Is it due to the dependencies? Is it that vcpkg is not available? Is it manually adding the `implot.h, implot_internal.h, implot.cpp, implot_items.cpp and optionally implot_demo.cpp` and you want to get them downloaded as part of cmake config step?

common use cases for proposing a CMake build for *ImPlot*:

1. *Dependency Management*
 I prefer to manage dependencies declaratively rather than manually copying files.
 When using *ImGui* and other libraries (e.g., GLFW, SDL2, OpenGL) that are already CMake-managed.

2. *No vcpkg or Package Manager*
 When working in an environment where *vcpkg* or other package managers (like Conan) aren't available or allowed.

3. *Automated Fetching via CMake*
  Want to use `FetchContent` or `ExternalProject_Add` to automatically download ImPlot during the CMake configuration step.
This avoids the need to manually clone or copy `implot.h`, `implot_internal.h`, `implot.cpp`, `implot_items.cpp`, and optionally `implot_demo.cpp`.

finnrorke:
Correct me if I'm wrong but ImGui does not have CMake support either. It is build tool agnostic. You can still use fetch content on this tool and ImGui. 

`FetchContent_Declare(
	imgui
	GIT_REPOSITORY https://github.com/ocornut/imgui.git
)`

`FetchContent_Declare(
	implot
	GIT_REPOSITORY https://github.com/epezent/implot.git
)`

`FetchContent_MakeAvailable(
        imgui
	implot
)`

`add_executable(${PROJECT_NAME} main.cpp
	${imgui_SOURCE_DIR}/imgui.cpp
	${imgui_SOURCE_DIR}/imgui_demo.cpp
    	${imgui_SOURCE_DIR}/imgui_draw.cpp
    	${imgui_SOURCE_DIR}/imgui_tables.cpp
    	${imgui_SOURCE_DIR}/imgui_widgets.cpp
[...]

PR #681, opened 2026-02-22
cmake: Add top-level CMakeList.txt

edowson:
This commit adds support for building an implot cmake library.


brenocq:
Thank you for the effort @edowson! As commented in https://github.com/epezent/implot/issues/667#issuecomment-3901261474, ImPlot tries to follow the ImGui standards as close as possible. Because of this, we won't support CMake until Cmake is supported by ImGui.

PR #683, opened 2026-02-25
A brand new manual, including ImGui, ImPlot and ImPlot3D!!!

pthom:
Hi @epezent, @brenocq (and @ocornut)

I spent quite some time in order to unify the manuals for Dear ImGui, ImPlot and ImPlot3D.

The idea is to have a common experience where a user can look at the demos to the left, and see their code to the right.

These days I have been in contact with Omar in order to finalize it. He is happy with the result and I am too.

**Urls for the manual**

* [Dear ImGui Manual with all 3 libs](https://pthom.github.io/imgui_manual): showing imgui, implot and implot3d (use the buttons in the top bar to switch to ImPlot)
* [Manual for ImGui](https://pthom.github.io/imgui_manual/?lib=imgui): showing only ImGui (it is the same url, with ?lib=imgui at the end)
* [Manual for ImPlot](https://pthom.github.io/imgui_manual/?lib=implot): showing only ImPlot (?lib=implot)
* [Manual for ImPlot3D](https://pthom.github.io/imgui_manual/?lib=implot3d): showing only ImPlot (?lib=implot3d)

**Notes about the manual**

- The source for the manual is now inside a [subfolder of imgui_bundle](https://github.com/pthom/imgui_bundle/tree/main/external/imgui_manual/imgui_manual) 
- ImGui Bundle is updated quite regularly, together with its dependencies (including imgui, implot, and implot3d)
- The manual displays the demo code in C++
- For Python users, it can also display the code in Python (the demo code is hosted in the bundle repo). Use the radio-buttons C++/Python. See below

image

---

**Changes required for the integration** (validated with Omar)

In order to integrate ImPlot to the manual, I had to make some small changes, which you can see in this PR. 
The principle was discussed and approved with Omar: see [this commit](https://github.com/ocornut/imgui/commit/8a15a1064d0b00d2628a20c3907da95b84fd34d4) where some tooling was added to support the IMGUI_DEMO_MARKER macro.

TLDR:
- This macro will call ImGui::DemoMarker (or do nothing on older ImGui)
- ImGui::DemoMarker will call a callback if registered, do nothing otherwise (so this is almost zero cost, or even zero once optimized)

(and the manual does register this callback of course)

---

**Update the old demo links**

You should update the old links to the demo in your repo and/or doc.
The old url https://traineq.org/implot_demo/src/implot_demo.html now redirects to https://pthom.github.io/imgui_manual 

Feel free to use the links that show only ImPlot or ImPlot3D. FYI, Omar will advertise the link that displays the 3 libs at the same time :-)

Cheers,

Pascal

### DIFF
--- a/implot_demo.cpp
+++ b/implot_demo.cpp
@@ -45,6 +45,14 @@
 
 #define CHECKBOX_FLAG(flags, flag) ImGui::CheckboxFlags(#flag, (unsigned int*)&flags, flag)
 
+// Helper to wire demo markers located in code to an interactive browser (e.g. imgui_explorer)
+#if IMGUI_VERSION_NUM >= 19263
+namespace ImGui { extern IMGUI_API void DemoMarker(const char* file, int line, const char* section); };
+#define IMGUI_DEMO_MARKER(section)  do { ImGui::DemoMarker("implot_demo.cpp", __LINE__, section); } while (0)
+#else
+#define IMGUI_DEMO_MARKER(section)
+#endif
+
 #if !defined(IMGUI_DISABLE_DEMO_WINDOWS)
 
 // Encapsulates examples for customizing ImPlot.
@@ -188,6 +196,7 @@ struct HugeTimeData {
 //-----------------------------------------------------------------------------
 
 void Demo_Help() {
+    IMGUI_DEMO_MARKER("Demo_Help");
     ImGui::Text("ABOUT THIS DEMO:");
     ImGui::BulletText("Sections below are demonstrating many aspects of the library.");
     ImGui::BulletText("The \"Tools\" menu above gives access to: Style Editors (ImPlot/ImGui)\n"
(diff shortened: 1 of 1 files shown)


pthom:
Hi @epezent  and @epezent (and @ocornut ) 

I am sorry if my message yesterday was a bit fast. 

Of course I am interested in your feedback concerning this and I want to wait for your it, and for your approval before putting it online! 

So in the meantime I restored the old implot and implot3d demo at 
https://traineq.org/implot_demo/src/implot_demo.html

You still can see the manuals in the urls I gave you, except that:

https://pthom.github.io/imgui_manual/ leads to a demo for ImGui only  
https://pthom.github.io/imgui_manual/?lib=all leads to a demo for all libraries
?lib=implot and ?lib=implot3d still work

(this way we wait until we all agree on whether the main demo should address the 3 libraries)


May be this thread will be a good place for us to discuss this (if you prefer to discuss it via other means, tell me :-)

PR #689, opened 2026-03-19
fix error: Bitwise operation between different enumeration types ('ImGuiButtonFlags_' and 'ImGuiButtonFlagsPrivate_') is deprecated

### LINKED ISSUE #688: [Bug] Bitwise operation between different enumeration types ('ImGuiButtonFlags_' and 'ImGuiButtonFlagsPrivate_') is deprecated
Recent updates to ImGui now cause build problems with ImPlot:

Image

Image

bostick:
Fixes https://github.com/epezent/implot/issues/688

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -216,6 +216,7 @@ You can read releases logs https://github.com/epezent/implot/releases for more d
 #if defined(__clang__)
 #pragma clang diagnostic ignored "-Wformat-nonliteral"  // warning: format string is not a string literal
 #pragma clang diagnostic ignored "-Wdeprecated-enum-enum-conversion" // warning: bitwise operation between different enumeration types ('XXXFlags_' and 'XXXFlagsPrivate_') is deprecated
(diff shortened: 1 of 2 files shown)


slowriot:
This doesn't "fix" anything, it just suppresses the warning.  See https://github.com/epezent/implot/pull/691 for an actual resolution.

bostick:
You are correct, but suppressing warning is how ImGui resolved the issue when it first appeared and I copied that approach.

https://github.com/sergeyn/imgui/commit/82bf0b684f34c1f55f4a0e0cf22dbd4250231f28

brenocq:
I prefer the solution implemented in ImGui (supress the warning) since it's cleaner (I don't personally like very verbose castings).

I did notice that the same error also happens in `implot_demo.cpp`, so I believe the warning suppression would also need to implemeneted there.

image

brenocq:
I think it should be all good now! Thank you @bostick!

PR #692, opened 2026-04-01
Fix: Harcoded colors

dkosmari:
This makes all color values through the `IM_RGB()` macro, to respect any non-default RGBA color ordering in ImGui, which is very necessary on big endian systems, and ImGui backends that use endian-dependent pixel formats.

I wrote this little Python script (`fix-colors.py`) to do the conversion:

```py
#!/bin/env python3

import fileinput

def convert_line(line):
    if len(line) == 0:
        return ''
    result = []
    tokens = [x.strip() for x in line.split(',')]
    for token in tokens:
        val = int(token)
        s = f'{val:08x}'
        a = int(s[0:2], 16)
        b = int(s[2:4], 16)
        g = int(s[4:6], 16)
        r = int(s[6:8], 16)
        result.append(f'IM_RGB({r},{g},{b})')

    return ','.join(result)

for line in fileinput.input():
    print(convert_line(line.strip()))
```

Here are the numbers extracted from `implot.cpp`: [numbers.txt](https://github.com/user-attachments/files/26417295/numbers.txt)

Then simply invoke the script above passing the `numbers.txt` as argument:

    ./fix-colors.py numbers.txt

It should print out the exact `IM_RGB()` sequences that are in the commit.


brenocq:
I noticed that the R and B channels are swapped

| Your PR | Expected | 
|----|----|
|image | image|

I do like the idea of using the defined IM_RGB instead of magic numbers, thank you!

dkosmari:
Oops, looks like the app I was using to test it had another endian-related problem and was swapping the R and B channels back.

PR #693, opened 2026-04-03
Fixes usage of mixed enum flags.

dkosmari:
- Undo warning suppression introduced in 9c34c33.
- ImGui's enums are `static_cast` to eliminate the warning.
- ImPlot's enums are extended with `_NoLegend` and `_NoFit` so there's never a need to mix them.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -215,8 +215,6 @@ You can read releases logs https://github.com/epezent/implot/releases for more d
 // Clang/GCC warnings with -Weverything
 #if defined(__clang__)
 #pragma clang diagnostic ignored "-Wformat-nonliteral"  // warning: format string is not a string literal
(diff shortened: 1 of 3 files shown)


brenocq:
I'm planning to keep the warning suppression since this was the [solution implemented in ImGui](https://github.com/ocornut/imgui/blob/49df3116bc5884f58031fd4bb8653e27e4044c96/imgui_widgets.cpp#L79-L80), when ImGui implements a different solution I'm more than happy to implement the same here.

I can see the benefit of extending ImPlot's enums with `_NoLegend`/`_NoFit`, but this makes it a bit harder to maintain when we add new item flags. For now I'll keep as is since the suppression was the ImGui solution.

dkosmari:
You have a `_None` flag repeated for every single one of these enums.

You already created maintenance hell because adding a new "generic" enum requires making sure it doesn't clash with any of the other "specialized" flags.

And in case somebody forgets to add the `_NoLegend` and `_NoFit` flags to a brand new enum, nothing will break, the user can still use the generic version (with an explicit cast if compiling with warnings.)

Issue #698, opened 2026-04-22
[Feature] A more capable skipper


EtienneParmentier:
**Context**
I am drawing a special kind of oscilloscope that is synchronized with the phase of my FM synthesizer.
While it make sense for me to draw lines between points that are near in phase; when my oscillator resets to zero it does not make sense to draw a line. Here is a screenshot with the offending lines:

<img width="1088" height="832" alt="Image" src="https://github.com/user-attachments/assets/a89925a2-7a84-44c3-ae9c-7dc43a4545ea" />

**Feature description**
I want to control which line segments should be skipped with a generic callback, like in ```ImPlot::PlotLineG```. I believe this is the most flexible approach. I could also add a flag but the behaviour would be very specific to my usecase.
I believe I have the bandwith and the skill to make a pull request, but I would like your opinion on the way to do this because I don't want to lose time on a design that would be rejected. For example, should I also implement this feature for other types of plots ?

I believe making yet another RendererLineXXX class is the way to go ?
Edit: I made a PR with my design: #699

EtienneParmentier:
PS: Please tell me yes explicitely I need clear communication

brmarkus:
> PS: Please tell me yes explicitely I need clear communication

My thumbs-up was just a reaction to your creative application and the inspirational feature.

EtienneParmentier:
Wait actually no, you do need to create another instance of all the line renderers !

EtienneParmentier:
Hm, It look like a custom plotter might be better suited to my needs. I believe adding filtering to the API is useful though since it allows to use all of the already made plotters / renderers.

PR #699, opened 2026-04-23
Feat: User defined line segments filters

EtienneParmentier:
Solves #698 .
Here is a picture to be more explicit:
Capture d’écran 2026-04-23 173235
I have not tested all flags combination, I believe adding an example in the demo would help ?
Also I have no idea what to do with markers since their are not line segments, should we tell the user with an assert ?

### DIFF
--- a/implot.h
+++ b/implot.h
@@ -992,7 +992,7 @@ IMPLOT_API void SetNextAxesToFit();
 IMPLOT_TMP void PlotLine(const char* label_id, const T* values, int count, double xscale=1, double xstart=0, const ImPlotSpec& spec=ImPlotSpec());
 IMPLOT_TMP void PlotLine(const char* label_id, const T* xs, const T* ys, int count, const ImPlotSpec& spec=ImPlotSpec());
 IMPLOT_API void PlotLineG(const char* label_id, ImPlotGetter getter, void* data, int count, const ImPlotSpec& spec=ImPlotSpec());
-IMPLOT_API void PlotLineFilter(const char* label_id, ImPlotGetter getter, ImPlotFilter filter, void* data, int count, const ImPlotSpec& spec=ImPlotSpec());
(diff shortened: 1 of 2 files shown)


brenocq:
Hi @EtienneParmentier — thanks a lot for the contribution, and for taking the time to write up a clear issue and a working prototype! It's great to see you getting comfortable navigating the ImPlot internals, and I hope you'll stick around to help develop new features down the line. 🙂

That said, I don't think this filtering belongs in ImPlot itself. We try to keep the ImPlot API as small and focused as possible, and extensions that are specific to a particular application are generally better handled on the user side. Segment filtering falls into that category.

The good news is your use case is already fully achievable with the current API. `PlotLine` renders a `NaN` point as a gap by default (`ImPlotLineFlags_SkipNaN` is actually the flag to *opt out* of that). So you can break the line exactly where you want by inserting a `NaN` between the points — for your oscilloscope, wherever the phase resets:

```cpp
// Rebuild only when the waveform changes, not every frame:
std::vector<ImPlotPoint> display;
display.reserve(N + 64);
for (int i = 0; i < N; ++i) {
    if (i > 0 && raw[i].x < raw[i-1].x)   // phase reset -> break the line here
        display.push_back({ NAN, NAN });
    display.push_back(raw[i]);
}

// Each frame — plain PlotLine, do NOT set SkipNaN:
ImPlot::PlotLine("signal", &display[0].x, &display[0].y,
                 (int)display.size(), 0, 0, sizeof(ImPlotPoint));
```

If you'd rather not keep a second buffer, the same idea works with `PlotLineG` by returning a `NaN` point for the break slots.

Let me know if this works for your application! And thanks again for the effort — really appreciate it.

EtienneParmentier:
Thanks for your help ! You have guessed my intent: providing a way to filter on the fly (without allocating). I honestly didn't thought about re-purposing the nan filters. However I believe I can solve the non allocating property by moving the nan insertion elsewhere in the code: I have to copy the data between two buffers at some point, so adding the null insertion code there would solve my line segment problems and my zero allocation requirements.

PR #702, opened 2026-04-30
Feat: Add Vertical and Horizontal markers

bruno-f-cruz:
This PR adds two new markers to the ImPlot library that are commonly used in other plotting libraries (see here: https://matplotlib.org/stable/gallery/lines_bars_and_markers/marker_reference.html)

These are useful for plots that try to cram a lot of events in a small space (for instance, raster plots: 
image)

It should be noted that while this appearance could already be achieved using simple lines, this makes the API easier to use and compatible with the scatter plots.

{46ADBC27-FD53-477B-8E5D-BDB23B68A3AA}


brenocq (approved):
Great work! Thanks for the minimal changes :)

I've modified the markers demo to centralize the texts/markers

PR #703, opened 2026-05-07
Fixes for ImDrawList::AddRect() signature change in 19276.

ocornut:
Fix for breaking change https://github.com/ocornut/imgui/commit/6df50a0667a35a853da38cf4a4cbcf632325fec9#diff-cbc47f9fc55eb2c19b4be92d1c1e137bb8b72d072588ab3cb4bcba88576d14cb

ImPlot3D would need a similar fix is using the last two parameters of `ImDrawList::AddRect()`.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -3171,8 +3171,13 @@ void EndPlot() {
     }
 
     // render border
+#if IMGUI_VERSION_NUM < 19276
     if (render_border)
-        DrawList.AddRect(plot.PlotRect.Min, plot.PlotRect.Max, GetStyleColorU32(ImPlotCol_PlotBorder), 0, ImDrawFlags_RoundCornersAll, gp.Style.PlotBorderSize);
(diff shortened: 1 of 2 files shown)


ocornut:
To clarify this is required to safely build with `#define IMGUI_DISABLE_OBSOLETE_FUNCTIONS`.

When Evan was away I used to merge/push occasionally hot-fixes directly myself.
@brenocq Would you like me to keep doing so or would you prefer that I let know merge the PR yourself?

(only asking for hot-fix relative to catching with imgui core changes)

brenocq:
Thanks for implementing the fix @ocornut!

> @brenocq Would you like me to keep doing so or would you prefer that I let know merge the PR yourself?

I would prefer for you to continue opening PRs so I get notified of the changes. This way I can also make sure I implement the necessary changes in ImPlot3D.

brenocq (approved):

ocornut:
Regardless I would open PR for visibility but I could potentially merge the hot fix for ImPlot and you still get notified. I’ll try to add ImPlot3d to my test bed so it’s better tested on my end too. Thanks for merging!

neondev9:
>  I’ll try to add ImPlot3d to my test bed so it’s better tested on my end too.

@ocornut  fyi: https://github.com/epezent/implot/discussions/694#discussioncomment-16506938

PR #706, opened 2026-05-20
Feat: Ignore `deprecated-enum-enum-conversion` warnings in GCC

samangh:
The fix to issue #688 was to ignore  `deprecated-enum-enum-conversion` warnings (see PR #689).

However, that pull request only implemented the fix for clang. This PR also implements this in GCC.


brenocq (approved):
Thanks for taking the time to also suppress the warning in GCC! LGTM

PR #711, opened 2026-08-04
Refactor: Replace horizontal/vertical AddLine() calls to AddLineH(),AddLineV()

ocornut:
Note: I do not suggest merging as-is. Read on.

[1.92.8](https://github.com/ocornut/imgui/releases/tag/v1.92.8) introduced helper `AddLineV()`, `AddLineH()` functions.
There have a nicer API but in current version is essentially identical.

An upcoming rework of ImDrawList will slightly change `AddLine()` (with a legacy path available) and basically using AddLineV()/AddLineH() will be both faster and a "simpler" path to backward compatibility.

This PR changes all horizontal/vertical `AddLine()` calls to `AddLineV`/`AddLineH`. 
There are only 3 remaining `AddLine()` which may need revaluation when the 1.93.0 is ready (either add +0.5 to provide pixel center, either pass ImDrawFlags_StrokeLegacy. Will provide details later as this is not out yet).

In theory you could merge this today and it makes ImPlot code a little bit simpler and shorter.
BUT, this requires 1.92.8, and unfortunately as `ImDrawList` functions are methods we can't easily add them locally.

I see 3 options:
- A. Merge as is and decide to break support for <1.92.8, which can be an occasion to cleanup some old support. This is sort of the lazy option.
- B. Make those helpers floating functions, e.g. `ImPlot_AddLineH(drawlist, ...` which makes it easy to support <1.92.8. I can change the PR to do that. It only creates a syntax asymmetry calling those functions.
- C. Use hacky/messy derived type ImDrawListForImPlot to add a method.

I would suggest (B) so it feels silly that the current PR is for (A) but I'll let you confirm what you think is the right approach.

### DIFF
--- a/implot.cpp
+++ b/implot.cpp
@@ -1833,9 +1833,9 @@ static inline void RenderGridLinesX(ImDrawList& DrawList, const ImPlotTicker& ti
             continue;
         if (xt.Level == 0) {
             if (xt.Major)
-                DrawList.AddLine(ImVec2(xt.PixelPos, rect.Min.y), ImVec2(xt.PixelPos, rect.Max.y), col_maj, size_maj);
+                AddLineV(&DrawList, xt.PixelPos, rect.Min.y, rect.Max.y, col_maj, size_maj);
             else if (density < 0.2f)
-                DrawList.AddLine(ImVec2(xt.PixelPos, rect.Min.y), ImVec2(xt.PixelPos, rect.Max.y), col_min, size_min);
+                AddLineV(&DrawList, xt.PixelPos, rect.Min.y, rect.Max.y, col_min, size_min);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


brenocq:
Thanks for making the changes! The new ImDrawList results look great, also, nice work on the Wiki! I agree (B) sounds like the best. I noticed the "Allow edits from maintainers" is turned off, so feel free to implement the new methods or allow me to edit the branch, then I can make the changes myself. I think it's always nice to keep support for old ImGui versions, when the code gets too messy then we can decide to drop support for ImGui < X.YYY.Z.

To offload some work from you, I could also start porting your ImPlot changes to ImPlot3D and add you as reviewer there, let me know if you prefer implementing the changes in both ImPlot and ImPlot3D yourself.

Thanks for the great work!

ocornut:
Thanks I’ll makes the change soon. Feel free to work on equivalent ImPlot3d change.

ocornut:
Reworked as 0182a86.
The signature and location of the helper functions made to match existing e.g.`AddTextVertical().`

brenocq (approved):
Awesome! I've tested it with `v1.92.4` and `features/drawlist_193`, both worked without problems.

brenocq:
@ocornut I've created the PR in ImPlot3D https://github.com/brenocq/implot3d/pull/195, feel free to take a look at it when you have some time (I'm not very familiar with AA lines).

PR #712, opened 2026-08-06
Refactor: update to upcoming v1.93 changes

ocornut:
Reference: https://github.com/ocornut/imgui/issues/9504
Without this lines would not be visible in new code as the layout of TexUvLines[] has changed.

Note that this PR includes the commit for #711


brenocq (approved):
Tested with ImGui `v1.92.4` and `features/drawlist_193`, all good. Thanks!

PR #721, opened 2026-09-18
Fix for calling two parameter version of ImGuiTextFilter::Draw()

ocornut:
Removed in https://github.com/ocornut/imgui/commit/a14bebd948e021a7801a9d987f135332dc514519
`SetNextItemWidth()` is available since 1.70, so no need for version check.


bostick:
Ran into this as well.

brenocq (approved):
Thanks for the PR! Fix also implemented in ImPlot3D: https://github.com/brenocq/implot3d/pull/196

